PR #1, opened 2016-04-11
Reformatting, new test

chris-smith:
-Basic tests for message generation
-Discovered some very strange formatting courtesy of changing Atom settings partway through (unfortunately that is most of this change)

### DIFF
--- a/example.js
+++ b/example.js
@@ -1,29 +1,33 @@
 'use strict';
 
 let rosjs = require('./index.js');
+const std_msgs = rosjs.require('std_msgs').msg;
+const SetBool = rosjs.require('std_srvs').srv.SetBool;
 
 rosjs.initNode('/my_node')
 .then((rosNode) => {
-  console.log('GOTIT');
   // EXP 1) Service Server
   let service = rosNode.advertiseService({
-    service: '/list_cameras',
-    type: 'baxter_core_msgs/ListCameras'
+    service: '/set_bool',
+    type: 'std_srvs/SetBool'
   }, (req, resp) => {
     console.log('Handling request! ' + JSON.stringify(req));
-    resp.cameras = ['right_hand_camera', 'left_hand_camera', 'head_camera'];
+    resp.success = !req.data;
+    resp.message = 'Inverted!';
     return true;
   });
 
   // EXP 2) Service Client
   let serviceClient = rosNode.serviceClient({
-    service: '/list_cameras',
-    type: 'baxter_core_msgs/ListCameras'
+    service: '/set_bool',
+    type: 'std_srvs/SetBool'
   });
   rosNode.waitForService(serviceClient.getService(), 2000)
   .then((available) => {
     if (available) {
-      serviceClient.call({}, (resp) => {
+      const request = new SetBool.Request();
+      request.data = true;
+      serviceClient.call(request, (resp) => {
         console.log('Service response ' + JSON.stringify(resp));
       });
     }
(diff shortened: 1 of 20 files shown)


example.js
```diff
@@ -1,10 +1,11 @@
 'use strict';
 
 let rosjs = require('./index.js');
+const std_msgs = rosjs.requires('std_msgs').msg;
+const ListCameras = rosjs.requires('baxter_core_msgs').srv.ListCameras;
```
IanTheEngineer: Do you really want your test to depend on `baxter_core_msgs`? Perhaps something similar to the `add_two_ints` service example?

utils/message_utils.js
```diff
@@ -26,57 +26,75 @@ let cmakePaths = cmakePath.split(':');
...
 
 let messagePackageMap = {};
+let messagePackagePathMap = {};
 
 //-----------------------------------------------------------------------
 // Utilities for loading, finding handlers for
 // message serialization/deserialization
 //-----------------------------------------------------------------------
 
 let MessageUtils = {
```
IanTheEngineer, line 41: Can you give a brief explanation (in comments) of how the message finding / caching / loading works for rosjs?

utils/network_utils.js
```diff
@@ -25,104 +25,104 @@ let MIN_PORT = 49152;
...
+
+        ipAddress = iface.address;
+        return true;
+      });
+    });
+    return ipAddress;
+  },
+
+  /**
    * FIXME: should this just return ROS_HOSTNAME
```
IanTheEngineer, line 68: Should this FIXME for ROS_HOSTNAME be an issue?

utils/network_utils.js
```diff
@@ -25,104 +25,104 @@ let MIN_PORT = 49152;
...
+    let _freePortCheck = (callback) => {
+      let port = getRandomPort();
+      portscanner.checkPortStatus(port, '127.0.0.1', (err, status) => {
+        // if the port is 'closed' then its not in use
+        if (status === 'closed') {
+          callback(port);
+          return;
+        }
+        //else
+        _freePortCheck(minPort, maxPort, callback);
```
IanTheEngineer, line 101: Is there any timeout for this `freePortCheck()` or is it infinitely recursive?

utils/serialization_utils.js
```diff
@@ -30,22 +30,22 @@ const Transform = require('stream').Transform;
...
+  this._messageLen = -1;
 
-	// as bytes of this message arrive, store them in this
-	// buffer until we have the whole thing
-	this._messageBuffer = [];
+  // as bytes of this message arrive, store them in this
+  // buffer until we have the whole thing
+  this._messageBuffer = [];
 
   // FIXME: These are specific to parsing a service response...
```
IanTheEngineer: Similarly, this should be an issue.

IanTheEngineer:
Just minor comments above. Aside from those, this looks great! :shipit:

index.js
```diff
@@ -111,20 +113,29 @@ let Rosjs = {
...
     rosNode = new RosNode(nodeName, rosMasterUri);
     return new Promise((resolve, reject) => {
       _checkMasterHelper(resolve, 0);
+    })
+    .catch((err) => {
+      log.error('Error: ' + err);
     });
   },
 
+  requires(msgPackage) {
```
IanTheEngineer: This should be `require()` rather than `requires()`

example.js
chris-smith: switched to std_srvs/SetBool

utils/network_utils.js
chris-smith, line 68: I don't think so currently - whether to use this or not is an option you can set when you initialize the node.

utils/network_utils.js
chris-smith, line 101: Not currently - it's assuming there's enough space that it will find one 'reasonably' quickly. That's also partially why it chooses a random port in the range instead of successively incrementing where it's more likely two things could discover 'hey this port is free' and then both try to use it. We do cover the case of 'this port was free but now its not' but clashes would occur more frequently.

utils/serialization_utils.js
chris-smith: FIXME here is maybe misleading... its more like a TODO. DeserializeStream is being used for any TCPROS deserialization, but the scheme is slightly different for services than for everything else so this is mostly me saying "Maybe don't conflate these things"

index.js
chris-smith: Thank you for including this comment for posterity

PR #3, opened 2016-04-21
Fix messages from workspace overlays

chris-smith:
We were not properly handling workspace overlays when finding message packages.


IanTheEngineer:
This makes sense. If I overwrite a message in my own development environment, I would expect that rosjs would use the overwritten message. An example (typical) prefix path:
`CMAKE_PREFIX_PATH=/data/users/imcmahon/dev/test_ik_solvers/devel:/opt/ros/indigo`

PR #5, opened 2016-04-23
Script to flatten message directories

chris-smith:
Adds a new script that will copy all of your message dependencies into a specified location, flattening the
directory structure that might exist due to workspace overlays. It will also parse through your files and update their require paths to take advantage of this new directory structure so that messages don't need to find their dependencies from your CMAKE_PREFIX_PATH. This is an optional step though, simply running catkin_make or catkin_make install will still work.

To use the script, cd in rosjs and run
`$ npm run flatten -- <desired_output_directory>`

This is built on top of my current pull request to genjs
RethinkRobotics-opensource/genjs#2


IanTheEngineer:
+1 this is great

PR #11, opened 2016-05-08
Api update

chris-smith:
Updates node handle sub, pub, service api to more closely resemble other roscpp/rospy. When advertising/subscribing, can supply either a string `'std_msgs/String'` or the class from `rosjs.require('std_msgs')`

e.g.

```
const StringMsg = rosjs.require('std_msgs').msg.String;

rosjs.advertise('chatter', 'std_msgs/String');
rosjs.advertise('chatter', StringMsg);
```

Adds some basic pub sub tests. Separates out `gennodejs` tests from client library tests so that people don't need to pull `test_msgs` down in order to run the client tests.

### DIFF
--- a/example.js
+++ b/example.js
@@ -7,21 +7,16 @@ const SetBool = rosnodejs.require('std_srvs').srv.SetBool;
 rosnodejs.initNode('/my_node')
 .then((rosNode) => {
   // EXP 1) Service Server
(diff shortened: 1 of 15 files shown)


IanTheEngineer:
So this new api looks great, and likely what users will want. Can we support both the current and old API's in the future? That would make things slightly more flexible for users.

PR #15, opened 2016-05-18
Update search path to match gennodejs

chris-smith:
In https://github.com/RethinkRobotics-opensource/gennodejs/pull/7 we updated the location of the generated messages to make future collisions less likely. This updates `rosnodejs` accordingly.


IanTheEngineer:
Great! Thanks for this update

PR #16, opened 2016-05-19
Unifying API for on-the-fly and gennodejs messages

### LINKED ISSUE #14: Unify API for on-the-fly generated and pre-compiled message classed
See https://github.com/RethinkRobotics-opensource/rosnodejs/pull/12#issuecomment-219899631.

This is complicated slightly by the fact that pre-compiled message classes can be instantiated synchronously, whereas on-the-fly generation of message classes is asynchronous. So a unified interface may need to follow the asynchronous approach, i.e., use a callback rather than a return value.

chfritz:
- returning promise from 'use'
- moved call to use into initNode

Unified APIs: rosnodejs.message now replaced with reosnodejs.require

Addresses part of #14.
- the way require now works is by first checking the registry of
  on-the-fly generated messages. If it finds the class there, it uses
  it, otherwise it will check the filesystem for the pre-compiled
  classes generated by gennodejs

### DIFF
--- a/index.js
+++ b/index.js
@@ -88,7 +88,8 @@ let Rosnodejs = {
         return Promise.resolve(this.getNodeHandle());
       }
       // else
-      throw new Error('Unable to initialize node [' + nodeName + '] - node [' + rosNode.getNodeName() + '] already exists');
+      throw new Error('Unable to initialize node [' + nodeName + '] - node [' 
+                      + rosNode.getNodeName() + '] already exists');
     }
 
     // FIXME: validate nodeName -- MUST START WITH '/'
@@ -119,8 +120,11 @@ let Rosnodejs = {
     // resolve when connection to master is established
     let checkMasterTimeout =  0;
     rosNode = new RosNode(nodeName, rosMasterUri);
+
     return new Promise((resolve, reject) => {
-      _checkMasterHelper(resolve, 0);
+      this.use(options.messages, options.services).then(() => {
+        _checkMasterHelper(resolve, 0);
+      });
     })
     .catch((err) => {
       log.error('Error: ' + err);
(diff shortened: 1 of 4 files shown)


chfritz:
Fixed the conflict.

index.js
```diff
@@ -181,17 +197,17 @@ let Rosnodejs = {
    * from .msg file if it doesn't already exist. This is mostly
    * because that would need to be async right now and we want a sync
    * method. */
-  message(type) {
-    return messages.getFromRegistry(type, "message");
-  },
+  // message(type) {
```
chris-smith: Any particular reason to keep this rather than pulling it?

index.js
```diff
@@ -139,39 +150,44 @@ let Rosnodejs = {
...
+          return self._useServices(services);
+        }).then(() => {
+          resolve();
+        });
     });
   },
 
   /** create message classes for all the given types */
-  _useMessages(types, callback) {
+  _useMessages(types) {
```
chris-smith, line 167: can you update this and _useServices to handle the case where `types` is undefined? People may not provide those if they're using compiled messages.

chris-smith:
This looks good! We should probably keep the bug open though because I believe there are still some differences to work out with respect to constructing messages (mostly with complex/nested types). I think some of that will require changes to `gennodejs` too.

chfritz:
OK, done. Also removed the special "fixed #xx"  term to avoid auto-closing the issue.

index.js
```diff
@@ -139,60 +150,54 @@ let Rosnodejs = {
...
-      console.log('get message type! ' + type);
-      messages.getMessage(type, function(error, Message) {
-        console.log('got it!');
-        Messages.push(Message);
-        if (Messages.length === types.length) {
-          callback();
-        }
+  _useMessages(types) {
+    if (!types || types.length == 0) {
+      resolve();
```
chris-smith: `return Promise.resolve()`

index.js
```diff
@@ -139,60 +150,54 @@ let Rosnodejs = {
...
       });
     });
   },
-
+  
   /** create message classes for all the given types */
-  _useServices(types, callback) {
+  _useServices(types) {
+    if (!types || types.length == 0) {
+      resolve();
```
chris-smith: `return Promise.resolve()`

chris-smith:
Tests and example run fine for me after `Promise.resolve` change - the service doesn't seem to be working in the on-the-fly example though. Could be an issue from the api update?

chfritz:
Thanks for the fix. Yes, seems like I shouldn't be coding when tired. Also found the problem with the service -- another result of my sloppy merging. Should be good now.

chris-smith:
No worries. Looks good now - thanks!

PR #18, opened 2016-05-24
Provide option to anonymize node

chris-smith:
- Adds support for `anonymous` option when calling `initNode` to anonymize your node name. Follows rospy logic.
- Fixes error message when unable to find message package.
- Also some unintended whitespace changes? Maybe from Atom?

e.g.

```
rospy.initNode('/my_node', {anonymous: true})
.then((nh) => {
   console.log( nh.getNodeName() ); // "/my_node_25878_1464120320242"
});
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -22,6 +22,7 @@
 const netUtils = require('./utils/network_utils.js');
 const msgUtils = require('./utils/message_utils.js');
 const messages = require('./utils/messages.js');
+const util = require('util');
 msgUtils.findMessageFiles();
 
 // these will be modules, they depend on logger which isn't initialized yet
@@ -72,6 +73,16 @@ function _validateNodeName(nodeName) {
   return nodeName;
 }
 
+/**
+ * Appends a random string of numeric characters to the end
+ * of the node name. Follows rospy logic.
+ * @param nodeName {string} string to anonymize
+ * @return {string} anonymized nodeName
+ */
+function _anonymizeNodeName(nodeName) {
+  return util.format('%s_%s_%s', nodeName, process.pid, Date.now());
+}
+
 let Rosnodejs = {
   /**
    * Initializes a ros node for this process. Only one ros node can exist per process
@@ -81,6 +92,12 @@ let Rosnodejs = {
    * @return {Promise} resolved when connection to master is established
    */
   initNode(nodeName, options) {
+    options = options || {};
+    if (options.anonymous) {
+      nodeName = _anonymizeNodeName(nodeName);
+      console.log('anonymous node name ' + nodeName);
+    }
+
     nodeName = _validateNodeName(nodeName);
 
     if (rosNode !== null) {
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -81,19 +92,23 @@ let Rosnodejs = {
...
       if (nodeName === rosNode.getNodeName()) {
         return Promise.resolve(this.getNodeHandle());
       }
       // else
-      throw new Error('Unable to initialize node [' + nodeName + '] - node [' 
+      throw new Error('Unable to initialize node [' + nodeName + '] - node ['
                       + rosNode.getNodeName() + '] already exists');
     }
 
-    // FIXME: validate nodeName -- MUST START WITH '/'
```
IanTheEngineer, line 95: Isn't this FIXME still true? String validation of the node name hasn't changed.

chris-smith, line 95: I meant to remove it when I added the validation that exists. There is, admittedly, a lot more name validation/mapping that we need to do, though this exists for most ROS names and not specifically for node names.

utils/message_utils.js
```diff
@@ -184,7 +184,7 @@ let MessageUtils = {
   loadMessagePackage(msgPackage) {
     const indexPath = messagePackagePathMap[msgPackage];
     if (indexPath === undefined) {
-      throw new Error('Unable to find message package %s', msgPackage);
+      throw new Error('Unable to find message package ' + msgPackage);
```
IanTheEngineer, line 187: Is there a JS style guide or linter dictating this change? If not, we should probably choose one. I assume this change is marginally better that it was previous?

chris-smith, line 187: error's won't format strings for you like logging will - so in the log you ended up just seeing
`Unable to find message package %s`

index.js
```diff
@@ -81,19 +92,23 @@ let Rosnodejs = {
    * @return {Promise} resolved when connection to master is established
    */
   initNode(nodeName, options) {
+    options = options || {};
+    if (options.anonymous) {
+      nodeName = _anonymizeNodeName(nodeName);
+      console.log('anonymous node name ' + nodeName);
```
IanTheEngineer: Just an FYI, this is the only spot in the file we `console.log()`, or any logging at all for that matter. Will this be replaced by your `rosnodejs` logging in the future?

IanTheEngineer:
Looks great overall. I'm ready to merge as is.

index.js
chris-smith: whoops! testing - meant to pull this out

utils/message_utils.js
IanTheEngineer, line 187: Noted! Not a marginal change at all, then.

index.js
IanTheEngineer: It's not necessarily bad to have it in there, as the pid and time are not obvious to the user when starting the node.

chris-smith: Will people typically care that they see the node name when anonymizing though? `rospy` doesn't log anything. They can also ask for the name for the name of the node at any point.

```
console.log(rosnodejs.nh.getNodeName());
```

IanTheEngineer: True, that works. Let's remove it.

PR #19, opened 2016-05-26
Centralize TCPROS Server

chris-smith:
This creates a single TCPROS server on the node to handle subscriber and service client connections. Previously each publisher and service was creating their own. Connections are handed off to publishers and services according to the connection header.

### DIFF
--- a/lib/Publisher.js
+++ b/lib/Publisher.js
@@ -75,20 +75,13 @@ class Publisher extends EventEmitter {
 
     this._subClients = {};
 
-    this._port = 0;
-
-    this._server = null;
-
     this._messageHandler = options.typeClass;
 
     // messages published before this publisher
     // was registered will be held here
     this._msgQueue = [];
 
-    this._setupTcp()
-    .then(() => {
-      this._register();
-    });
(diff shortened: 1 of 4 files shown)


chris-smith:
@IanTheEngineer  can you look at this and merge/dissent?

lib/Publisher.js
```diff
@@ -213,110 +206,60 @@ class Publisher extends EventEmitter {
...
-      });
-    };
+    subscriber.on('end', () => {
+      this._log.info('Sub %s sent END', subscriber.name);
+    });
 
-    return new Promise((resolve, reject) => {
-      _createServer(resolve);
+    subscriber.on('error', () => {
+      this._log.warn('Sub %s had error', subscriber.name);
```
IanTheEngineer, line 251: Seems a bit funny to me that this error is reported as a warning. Was that intended?

IanTheEngineer:
This pull request looks great. Sorry it took me so long to review. I'm going to merge as-is. At some point we should document/comment the functions as to their intended purposes (or at least the public-facing API).

PR #25, opened 2016-06-18
Compute correct service md5sums and generate action messages

chfritz:
Note that this includes PR #23 as that is includes code that I here needed to refactor.

### DIFF
--- a/utils/messages.js
+++ b/utils/messages.js
@@ -23,19 +23,13 @@ messages.getPackageFromRegistry = function(packagename) {
 /** ensure the handler for this message type is in the registry,
  * create it if it doesn't exist */
 messages.getMessage = function(messageType, callback) {
-  getMessageFromPackage(messageType, ["msg"], callback);
+  getMessageFromPackage(messageType, "msg", callback);
 }
 
 /** ensure the handler for requests for this service type is in the
  * registry, create it if it doesn't exist */
-messages.getServiceRequest = function(messageType, callback) {
-  getMessageFromPackage(messageType, ["srv", "Request"], callback);
-}
-
-/** ensure the handler for responses for this service type is in the
- * registry, create it if it doesn't exist */
-messages.getServiceResponse = function(messageType, callback) {
-  getMessageFromPackage(messageType, ["srv", "Response"], callback);
+messages.getService = function(messageType, callback) {
+  getMessageFromPackage(messageType, "srv", callback);
 }
 
 // ---------------------------------------------------------
@@ -71,7 +65,7 @@ var registry = {};
    @param messageType is the ROS message or service type, e.g.
    'std_msgs/String'
    @param type is from the set 
-   [["msg"], ["srv","Request"], ["srv","Response"]]
+   [["msg"], ["srv","Request"], ["srv","Response"] 
 */
 function getMessageFromRegistry(messageType, type) {
   var packageName = getPackageNameFromMessageType(messageType);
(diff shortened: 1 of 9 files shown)


utils/messages.js
```diff
@@ -185,15 +195,54 @@ function parseMessageFile(fileName, details, type, callback) {
...
+                // we need to clone what's already there in details
+                // into the sub-objects
+                request: JSON.parse(JSON.stringify(details)),
+                response: JSON.parse(JSON.stringify(details))
+              };
+              rtv.request.constants = aggregate[0].constants;
+              rtv.request.fields = aggregate[0].fields;
+              rtv.response.constants = aggregate[1].constants;
+              rtv.response.fields = aggregate[1].fields;
+              rtv.request.md5 = rtv.response.md5 = calculateMD5(rtv, "srv");
```
chris-smith, line 200: I had been planning to make the service md5 accessible outside the request or response objects since its the hash of both of them. So the generated service file would export

```
{
Request: ServiceRequest,
Response: ServiceResponse,
md5() { // return service md5 sum }
}
```

That seems like it might be difficult for you to duplicate here though?

chris-smith:
All the new action parsing stuff looks good. I'm mostly wondering about how its exposed - with catkin, action message definitions are just turned into message files that, from the client library's perspective, are indistinguishable from regular messages. Importing the generated files from a `Move` action message in `rospy` would look like

```
from my_actions.msg import MoveGoal, MoveFeedback, MoveResult, MoveAction, MoveActionGoal, MoveActionFeedback, MoveActionResult
```

compared with importing a standard message

```
from std_msgs.msg import String
```

It's nice that they're exposed separately here, but I'm not sure that I have control over how that gets done in `gennodejs` so I don't know that I could follow this pattern.

utils/messages.js
chfritz, line 200: That should be fine. I would prefer having a proper service class as well, rather than using the Message class for everything. But I believe we'll need to coordinate this change, for our two class definitions in order to be able to also update where the md5 is being used, such as in ServiceClient: `this._messageHandler.Response.md5sum()`  would need to change to `this._messageHandler.md5sum()`. Right?

chfritz:
I see what you are saying. Let me refactor a little.

chfritz:
BTW, how come you guys have not yet run into this issue:
https://github.com/baalexander/node-xmlrpc/pull/132

Have you been able to control turtlesim (via /turtle1/cmd_vel) using rosnodejs? I was unable to do that before fixing the above in the xmlrpc library. The turtle simply wouldn't move: the tcpros message was never sent, because turtlesim never reacted to the method response of the requestTopic method call. That was because that method response didn't send a content-length header. Have you seen any such issues before?

With this out of the way, I should be able to fix up my action handling rather quickly (I'll basically just remove it, since users can use the action messages directly).

chfritz:
OK, refactored and rebased. This should be ready now. I've also added a new example that tests with turtlesim. To run, just run:

``` sh
rosrun turtlesim turtlesim_node
rosrun turtle_actionlib shape_server
node example_turtle.js
```

chris-smith:
Did you have a strange rebase possibly? This doesn't look it contains the action parsing changes to me - I'd expect something in `utils/messages.js` to call `getMessageFromPackage(messageType, "action" callback)`

chfritz:
I actually removed all of that once I realized that catkin will generate the action messages already. So the turtle example now uses those .msg files directly.

chris-smith:
Ah ok. This looks good to me, tests run fine too.

Issue #20, opened 2016-06-10
Update logging


chris-smith:
Logging right now is very rough. It should be cleaned up, provide similar options/interface to [`rospy`](http://wiki.ros.org/rospy/Overview/Logging) or [`roscpp`](http://wiki.ros.org/roscpp/Overview/Logging), log to `/rosout`, etc. I've started reworking it so that you could do something like

```
rosnodejs.logInfo("You should know about this");
rosnodejs.logInfoThrottle(1000, "You should know about this every so often");
rosnodejs.logInfoOnce("I'm not going to repeat myself");
```

or alternately ask for a logger and not need to pass around rosnodejs

```
let log = rosnodejs.createLogger("myCustomLogger");
log.info("Hey, that's cool");
```

I'm currently wrapping a [`bunyan`](https://github.com/trentm/node-bunyan) logger and and plan to expose some ability to add streams and modify formatting. Throttling right now is done by temporarily storing the first argument (excluding throttle time) in a set - this would make formatted strings strongly preferred over composed ones because:

```
let i = 0;
setInterval(() => {
  log.infoThrottle(1000, "Counter: %d", i);  // prints once a second
  log.infoThrottle(1000, "Counter: " + i);   // prints twice a second
  ++i;
}, 500);
```

Very open to recommendations about other ways to do this or anything else, but before diving into this too much more I wanted to at least publicly check if people (a) hate some of this, (b) hate all of this, or (c) have other opinions/thoughts.

chris-smith:
Nobody had anything to say, but this is in. We'll deal with issues as they come.

PR #26, opened 2016-06-18
Logging Update

chris-smith:
Pull request for issue #20. This is built on #19. I apparently also based it off of changes to add persistent service clients, which I thought was already in upstream but I guess I forgot to put up...
- Pulls in `bunyan` for logging backend.
- Exposes various log methods through `rosnodejs.log`.
- Publish logs to `/rosout`
- Logging tests


example.js
```diff
@@ -9,28 +9,34 @@ rosnodejs.initNode('/my_node')
   // EXP 1) Service Server
   let service = rosNode.advertiseService('/set_bool', SetBool,
     (req, resp) => {
-      console.log('Handling request! ' + JSON.stringify(req));
+      rosnodejs.log.info('Handling request! ' + JSON.stringify(req));
```
IanTheEngineer, line 12: In general, I am a fan of verbosity. And I do like this fact that it is super obvious the kind of logger you are using here. But I am wondering how easy would it be for a user to alias `rosnodejs.log.info` as something like `ros.info`? Not saying that we need to do it for them though, just curious.

IanTheEngineer:
Thank you for adding logging! Looks really good to me

chris-smith:
I'll squash this again before this gets merged - had a few fixes to make after pulling in upstream though and figured I'd make it clearer - should just be fd193d2.

example.js
chris-smith, line 12: I think we could provide that - `rospy` provides all the methods with a `log` prefix to them (`logerr`, `loginfo`, `logwarn`, etc). This method lets you pull the logger out yourself and never refer to `rosnodejs` again though.  So you could do

```
let log = rosnodejs.log;
log.info('A Message!');
```

You can also generate your owned named logger as a child of the root logger if you want to (and then it would be available through `/get_loggers` and  `/set_logger_level`)

```
let customLogger = rosnodejs.log.getLogger('MyLogger');
customLogger.info('My other message');
```

You're less likely to get  `throttle` or `once` clashes too if you create your own loggers as they each keep their own sets.
The `log` property from `rosnodejs` is actually the `LoggingManager` though.  It exposes the logging methods of its root logger for ease of use. So all the methods available through `rosnodejs.log` are not necessarily available through loggers that you ask it to create.

PR #29, opened 2016-07-08
Getting the action client to work, and required updates for on-demand messages

chfritz:
More closely matching behavior of gennodejs created classes
- default values for each primitive type
- now using actionclient in turtle example

Auto-requiring essential message types if not available by gennodejs (e.g., for logging)
- also polished the turtle example a little more

### DIFF
--- a/utils/fields.js
+++ b/utils/fields.js
@@ -1,23 +1,40 @@
+'use strict';
+
 var fields = exports;
 
-fields.primitiveTypes = [
-  'char'
-, 'byte'
-, 'bool'
-, 'int8'
-, 'uint8'
-, 'int16'
-, 'uint16'
-, 'int32'
-, 'uint32'
-, 'int64'
-, 'uint64'
-, 'float32'
-, 'float64'
-, 'string'
-, 'time'
-, 'duration'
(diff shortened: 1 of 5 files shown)


utils/fields.js
```diff
@@ -1,23 +1,34 @@
...
+  'float64': 0,
+  'string': '',
+  'time': 0,
+  'duration': 0
+};
+
+fields.primitiveTypes = Object.keys(map);
+
+fields.getDefaultValue = function(type) {
+  if (type.indexOf("[]") == type.length - 2) {
```
chris-smith: typically if this is a constant length array (e.g. `float64[9] K` in `sensor_msgs/CameraInfo`) it would be set here as well.

```
getDefaultValue ('float64[9]')
  => return new Array(9).fill(getDefaultValue('float64');
```

chris-smith:
Everything else looks good here - happy to merge once that's in or if there's a bigger change set required that I'm unaware of for that.

chfritz:
Great suggestion. Thanks! Done and squashed.

chris-smith:
Nice!

PR #31, opened 2016-08-29
Fixing an issue on node 4.5

chfritz:
- now using ES6 class with `extends` instead of prototype pattern with
  `util.inherits`. This works on both node 4.5 and newer ones, and is the recommended way now.


chris-smith:
This looks good to me - can you just confirm that the tests (`npm test`) still work? Keep meaning to get travis running those automatically...

chfritz:
Yes, all 28 tests are passing.

PR #34, opened 2016-10-15
Fix for prior fix.

chfritz:
Sorry, there was an issue with the previous attempt of fixing this.
This one seems to work.

### DIFF
--- a/utils/messages.js
+++ b/utils/messages.js
@@ -341,14 +341,27 @@ function extractFields(content, details, callback) {
 
         if (equalIndex !== -1) {
           fieldName = field.substring(0, equalIndex).trim();
+          // truncate comments and whitespace
           var constant = field.substring(equalIndex + 1, field.length).trim();
+          var constantEnd = constant.length;
+          var constantComment = constant.indexOf("#");
+          if (constantComment >= 0) {
+            if (fieldType != "string") {
+              var firstWhitespace = constant.match(/\s/);
+              if (firstWhitespace) {
+                constantEnd = firstWhitespace.index;
+              }
+            } else {
+              constantEnd = constantComment;
+            }
+          }
           var parsedConstant = fieldsUtil.parsePrimitive(fieldType, constant);
 
           constants.push({
             name        : fieldName
             , type        : fieldType
             , value       : parsedConstant
-            , raw         : constant
+            , raw         : constant.slice(0, constantEnd)
             , index       : fields.length
             , messageType : null
           });
@@ -504,7 +517,9 @@ function buildMessageClass(details) {
             new (field.messageType)(values ? values[field.name] : undefined);
         } else {
           // simple value
-          that[field.name] = values ? values[field.name] :
+          that[field.name] =
+            (values && typeof values[field.name] != "undefined") ? 
+            values[field.name] :
             (field.value || fieldsUtil.getDefaultValue(field.type));
         }
       });


chris-smith:
Did the previous commit just have an issue with comments? ROS seems to use the entire line after the equals sign (up to a comment if it's not a string) to calculate the md5 text - so (`float32 FLOAT_CONST=1.0` gives a different md5sum than `float32 FLOAT_CONST=1.00`)

I had to do something similar in the [indigo branch](https://github.com/chris-smith/rosjs/blob/indigo-devel/utils/messageGeneration/MessageSpec.js#L247) actually - ended up consulting `genmsg` source code a lot. Mainly [gentools](https://github.com/ros/genmsg/blob/490b3753213ed3e2251acd218c62224b2787ee0e/src/genmsg/gentools.py#L59), [msgs](https://github.com/ros/genmsg/blob/490b3753213ed3e2251acd218c62224b2787ee0e/src/genmsg/msgs.py#L158), and [msg_loader](https://github.com/ros/genmsg/blob/490b3753213ed3e2251acd218c62224b2787ee0e/src/genmsg/msg_loader.py#L183)

chfritz:
Thanks for the pointer. That is indeed a better way of dealing with the issue and I've adopted that now, but I still needed to store the constant value as a string, rather than a parsed value in order to stop javascript from doing things like turning 1.0 into 1.

jeffadamsc:
Do you still have a problem with  "1.0" vs "1.00" in this change? I see you're looking for whitespace, but not trailing zeros.

utils/messages.js
```diff
@@ -341,14 +341,31 @@ function extractFields(content, details, callback) {
...
+              constantEnd = constantComment;
+            }
+          }
           var parsedConstant = fieldsUtil.parsePrimitive(fieldType, constant);
 
           constants.push({
             name        : fieldName
             , type        : fieldType
             , value       : parsedConstant
+<<<<<<< HEAD
```
chris-smith: merge issue

utils/messages.js
```diff
@@ -504,8 +521,7 @@ function buildMessageClass(details) {
             new (field.messageType)(values ? values[field.name] : undefined);
         } else {
           // simple value
-          that[field.name] = values ? values[field.name] :
-            (field.value || fieldsUtil.getDefaultValue(field.type));
+          that[field.name] = values ? values[field.name] : null;
```
chris-smith: also a merge issue?

chris-smith:
> Do you still have a problem with "1.0" vs "1.00" in this change? I see you're looking for whitespace, but not trailing zeros.

1.0 is different from 1.00 in ros' message md5 text so trailing zeros are correct.

utils/messages.js
chfritz: oops, sorry. fixed now.

utils/messages.js
chfritz: no this one was intended.

chris-smith: Won't this force users to specify values for all the fields in their messages?

```
const header = new std_msgs.msg.Header()
pub.publish(header); // invalid
```

whereas before that would have worked

chfritz:
I'm actually not sure what the right behavior is. We also want to allow users to leave (optional) values unspecified (null). Maybe we just need to set them explicitly to null then already in the JS object.

chris-smith:
I don't think ROS 1.0's message serialization scheme allows for optional values.

If we want optional arguments when sending messages we typically will use constants or something and check against those on the other end.

```
if (msg.optionalField !== MyMessage.OPTIONAL_DEFAULT) { ... }
// or
if (msg.optionalList.length > 0) { ... }
```

It does start to get cumbersome when you want optional complex fields.

Automatically providing default values for fields allows devs sending a message to ignore certain fields if they don't care about them though - provided your subscriber interprets the default values correctly...

chfritz:
I guess you are right. I was thrown off by some null values I saw in our messages, but I guess those are just used as constants the same way you guys are using them.

Fixed that now, allowing for null values but providing defaults for fields that are actually `undefined`, i.e., missing in the goal object.

chris-smith (approved):

chris-smith:
Awesome - I'll release a new version with this and then pull the indigo merge changes in.

PR #35, opened 2016-10-19
Merge indigo/kinetic branches

chris-smith:
This is a sizable merge of my `indigo-devel` branch on to `kinetic-devel`.
-Adds ability to generate code similar to `gennodejs`
-Fixes a number of bugs
-Updates ServiceClient calls to queue, allowing multiple successive calls that will execute in order
-Updates xmlrpc master/slave code so that calls are queued/retried in order instead of every client retrying on their own
-Exposes access to Log streams for other code to use (if they have their own logging system and also want to log through ROS)
-Support `ROSCONSOLE_JS_FORMAT` environment variable to format console logging (similar to `ROSCONSOLE_FORMAT`, but we don't support filename/line #)
-Simplifies host name logic (`ROS_IP || ROS_HOSTNAME || os.hostname()`)
-More TCPROS Header validation
-More testing
-Moves examples of repo root

Ran `npm test`, message generation (`npm run generate`), message tests (`npm run gennodejsTests`) successfully.

I could merge this into an `indigo-devel` branch here but I think we'd want to merge them all into `kinetic-devel` anyway - the only difference should be that users working in `kinetic` should have pre-existing messages available to them.

Releasing a new version of `rosnodejs` off `kinetic-devel` would also be premised on a new release of `gennodejs` to support changes to the generated message code - it should be some additional changes to [this old pull request](https://github.com/RethinkRobotics-opensource/gennodejs/pull/12).

Before merging this in, I think we should probably tag a release of ~ the current implementation and possibly push to npm as this changes a lot.


IanTheEngineer:
For reference, the `ROSCONSOLE_FORMAT` syntax is outlined here: http://wiki.ros.org/rosconsole#Console_Output_Formatting

PR #40, opened 2016-11-13
Fix on demand default arrays + support for Node 4.5

chfritz:
1. Fixes an issue with arrays whose value is not specified during message creation, when using the onTheFly message definitions.

2. Counterpart to https://github.com/RethinkRobotics-opensource/ros_msg_utils/pull/6. Once the other PR is merged in ros_msg_utils, we can change the dependency back to the RethinkRobotics-opensource repo.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "rosnodejs",
   "version": "1.0.3",
   "description": "Native ROS for nodejs",
-  "main": "index.js",
+  "main": "dist/index.js",
   "keywords": [
     "ros"
   ],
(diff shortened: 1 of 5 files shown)


chris-smith:
Ah! Sorry I broke the default arrays - comments for babel here will be the same as those over on RethinkRobotics-opensource/ros_msg_utils#6

chfritz:
You didn't break anything. They were broken all along ;-)

package.json
```diff
@@ -24,12 +26,14 @@
     "mocha": "^2.4.5"
   },
   "dependencies": {
+    "babel-cli": "^6.18.0",
```
chris-smith: should these be `devDependencies`?

chfritz:
fixed. thanks.

package.json
```diff
@@ -21,15 +23,17 @@
...
+    "babel-preset-es2015": "^6.18.0"
   },
   "dependencies": {
     "argparse": "1.0.7",
     "async": "2.0.1",
     "bunyan": "1.8.1",
     "md5": "2.1.0",
     "moment": "2.12.0",
-    "ros_msg_utils": "RethinkRobotics-opensource/ros_msg_utils#v1.0.1",
+    "ros_msg_utils": "chfritz/ros_msg_utils#add_babel",
```
chris-smith: `ros_msg_utils` has been published so you should be able to set this to 1.0.2 - I think that's the only thing left to update

chfritz:
great! done.

PR #41, opened 2016-12-01
Full Message Definition, Fix ros bags

chris-smith:
- Calculates the full message definition for generated code.
- Includes message definition in publisher, subscriber connection headers
- tests for message definitions
- adds back shutdown call that was lost somehow

Bagged complicated message - fixed, played, opened with `rqt_bag`
`gennodejs` should already calculate the full message definition so I believe its messages will now also work for ros bags (we should test this).


IanTheEngineer:
Superseded by #48

PR #43, opened 2016-12-12
Fix tests

chris-smith:
Issues with how tests were interacting with the ros master and each other were causing some of them to break. Creates a cheap master stub that, for now, only provides `getUri`. Tests are passing again.


IanTheEngineer:
Looks great! Thank you for fixing the tests.

Issue #45, opened 2017-01-19
Trivial program errors continuously


cinderblock:
I'm trying to run this very trivial node program:

```
const rosnodejs = require('rosnodejs');
const std_msgs = rosnodejs.require('std_msgs');

rosnodejs.initNode('my_node')
.then(nh => {
  const pub = nh.advertise('/topic', std_msgs.msg.String);
  const msg = new std_msgs.msg.String();
  msg.data = 'Hello!';
  setInterval(() => {
    pub.publish(msg);
  }, 1000);
});
```

Using the `kinetic-devel` git branch.

It connects to master just fine but then instantly fills the screen with repeated error messages.

```
[INFO] [1484793140.047] (ros): Connected to master at http://localhost:11311!
[ERROR] [1484793140.057] (ros.rosnodejs): Exception trying to advertise service /my_node/get_loggers
[WARN] [1484793140.058] (ros): Unable to setup ros logging services
[ERROR] [1484793140.063] (ros.rosnodejs): Error when publishing message on topic /rosout: TypeError: MessageClass.getMessageSize is not a function
    at Object.serializeMessage (/home/cameron/ros/node-test/node_modules/rosnodejs/utils/tcpros_utils.js:264:34)
    at msgQueue.forEach (/home/cameron/ros/node-test/node_modules/rosnodejs/lib/Publisher.js:168:43)
    at Array.forEach (native)
    at Publisher._handleMsgQueue (/home/cameron/ros/node-test/node_modules/rosnodejs/lib/Publisher.js:162:14)
    at ClientQueue.handleClientMessages (/home/cameron/ros/node-test/node_modules/rosnodejs/utils/GlobalSpinner.js:45:22)
    at GlobalSpinner._handleQueue (/home/cameron/ros/node-test/node_modules/rosnodejs/utils/GlobalSpinner.js:203:24)
    at ontimeout (timers.js:365:14)
    at tryOnTimeout (timers.js:237:5)
    at Timer.listOnTimeout (timers.js:207:5)
[ERROR] [1484793140.063] (ros.rosnodejs): Error when publishing message on topic /rosout: TypeError:
...
```

I'm a little new to ROS still. The first error & warning looks like it's just that I don't have a ROS logger running.

I'm at a loss as to what the other errors are being caused by. Maybe that would be fixed by logger stuff. I haven't had the time to try it yet.

Any insight would be greatly appreciated.

chris-smith:
[...]

PR #46, opened 2017-01-21
add datatype function to generated message class

chris-smith:
Fixes an issue discovered in #45 when using the message class in advertise, subscribe calls


IanTheEngineer:
Superceded by #48

PR #47, opened 2017-01-24
travis ci integration

rethink-neil:
create minimal `.travis.yml`, add badge to `README.md`

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,1 +1,1 @@
-# rosnodejs
+# rosnodejs [![Build Status](https://travis-ci.org/RethinkRobotics-opensource/rosnodejs.svg)](https://travis-ci.org/RethinkRobotics-opensource/rosnodejs)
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -1 +1 @@
-# rosnodejs
\ No newline at end of file
+# rosnodejs [![Build Status](https://travis-ci.org/rethink-neil/rosnodejs.svg)](https://travis-ci.org/rethink-neil/rosnodejs)
```
IanTheEngineer: We should set this up under the @RethinkRobotics-opensource group user.

rethink-neil: fixed, changed `rethink-neil` to `RethinkRobotics-opensource`

chris-smith:
Looks good to me. We should figure out the best way to hook this up to our actual tests... probably involves some amount of `ros` installation? Though I don't think `travis` supports 16.04 so we can't test against `kinetic` without running through tests through docker. We could also skip the `ros` install and just setup the minimal environment we need manually.

PR #48, opened 2017-02-01
Integrates ros_msg_utils & adds basic Time

### LINKED ISSUE #36: Better namespace support
ROS supports a lot of namespacing options (base, relative, global, and private) and validation that we don't provide. We should. `NodeHandle`'s should provide quick access to nested resources.
```
const nh = rosnodejs.nh('/my/local/namespace');
const sub = nh.subscribe('input_data'); // subscribe to /my/local/namespace/input_data
```

### LINKED ISSUE #38: Fix rosbag recording
[`rosbag`](http://wiki.ros.org/rosbag) seems to depend on publishers including the message definition in their connection header although that field is not mentioned in the [spec](http://wiki.ros.org/ROS/TCPROS). Without the message definition, `rosbag` tools complain about mismatching md5 sums.

Including the message definition still results in different bags, though I'm not certain if this is a problem.

Message definition in connection header for a `rosnodejs` bag
```
message_definition=# A Pose with reference coordinate frame and timestamp
    Header header
    Pose pose
```

Message definition in connection header for a `rospy` bag
```
message_definition=# A Pose with reference coordinate frame and timestamp
Header header
Pose pose

================================================================================
MSG: std_msgs/Header
# Standard metadata for higher-level stamped data types.
# This is generally used to communicate timestamped data 
# in a particular coordinate frame.
# 
# sequence ID: consecutively increasing ID 
uint32 seq
#Two-integer timestamp that is expressed as:
# * stamp.sec: seconds (stamp_secs) since epoch (in Python the variable is called 'secs')
# * stamp.nsec: nanoseconds since stamp_secs (in Python the variable is called 'nsecs')
# time-handling sugar is provided by the client library
time stamp
#Frame this data is associated with
#0: no frame
#1: global frame
string frame_id

================================================================================
MSG: geometry_msgs/Pose
# A representation of pose in free space, composed of postion and orientation. 
Point position
Quaternion orientation

================================================================================
MSG: geometry_msgs/Point
# This contains the position of a point in free space
float64 x
float64 y
float64 z

================================================================================
MSG: geometry_msgs/Quaternion
# This represents an orientation in free space in quaternion form.

float64 x
float64 y
float64 z
float64 w

```

chris-smith:
Each of these is roughly organized into its own commit

-migrates all source code into `src/`
-moves `ros_msg_utils` into `rosnodejs` from its separate repo
-adopts `bn.js` for 64-bit integers. I found `bn.js` to be more flexible than `bignum` and has better performance.
-Adds a basic Time interface for getting current (or sim) time (`rosnodejs.Time.now()` will return sim time if appropriate), converting between `Date` and `ros time` representations.
-Use full message definition in publisher connection headers. This fixes an issue where you could not play back `rosbags` from `rosnodejs`. As part of this, I changed some of how on the fly messages are generated to share infrastructure with message generation in order to compute the full message definition. It also changes the on the fly message signature to more closely align with the generated ones - I'm open to putting some of this back. (Resolves #38)
-Adds the ability to namespace through NodeHandles (Resolves #36)


src/lib/Time.js
```diff
@@ -0,0 +1,46 @@
...
+
+let simTimeSub = null;
+let simTime = timeUtils.dateToRosTime(0);
+
+function handleSimTimeMessage(msg) {
+  console.log('Got sim time update: %j', msg);
+  simTime = msg.clock;
+}
+
+const Time = {
```
IanTheEngineer: We should create an Issue to track the fact that Sim time will not be represented in any of the logs (they will use the wall-clock time).

src/utils/messageGeneration/packages.js
```diff
@@ -198,10 +283,9 @@ exports.findPackage = function(packageName, callback) {
     callback(null, directory);
     return;
   }
-  var rosRoot = process.env.ROS_ROOT;
-  var packagePath = process.env.ROS_PACKAGE_PATH
-  var rosPackagePaths = packagePath.split(':')
-  var directories = [rosRoot].concat(rosPackagePaths);
+  var packagePath = process.env.ROS_PACKAGE_PATH;
```
IanTheEngineer: Are any checks needed to report when the `ROS_PACKAGE_PATH` isn't set?

src/lib/NodeHandle.js
```diff
@@ -19,15 +19,22 @@
...
+const namespaceUtils = require('../utils/namespace_utils.js');
 const ActionClient = require('./ActionClient');
 
 class NodeHandle {
-  constructor(node) {
-    if (!(node instanceof RosNode)) {
-      throw new Error('Unable to create NodeHandle from type ' + typeof(node));
-    }
+  constructor(node, namespace=null) {
+    // if (!(node instanceof RosNode)) {
```
IanTheEngineer: Womp womp. Remove dead code please.

IanTheEngineer (approved):
Lots of solid improvements here. Thanks for the update! Be sure to grab some engine folk to double-check before merging.

IanTheEngineer:
@chfritz would you give this a quick review, please? These changes touch a fair amount of your code

chfritz:
Will do. Should be able to find the time tonight or over the weekend.

chfritz:
Looks good to me. But there are minor changes I had to make to the turtlesim example to make it work again with the new code. I've created a PR for that into this branch: https://github.com/chris-smith/rosjs/pull/10. Please consider merging that before merging this one into kinetic-devel.

Great work consolidating things, @chris-smith !

chris-smith:
Thanks for catching that! Merged!

PR #49, opened 2017-02-17
Fix no goal_id bug and no header bug

mjyc:
Generate a GoalID or Header message when the goal_id or the header field is missing from the input goalMessage.

### DIFF
--- a/src/lib/ActionClient.js
+++ b/src/lib/ActionClient.js
@@ -21,6 +21,7 @@ const timeUtils = require('../utils/time_utils.js');
 const msgUtils = require('../utils/message_utils.js');
 const EventEmitter = require('events');
 let GoalID = null;
+let Header = null;
 
 class ActionClient extends EventEmitter {
   constructor(options) {
@@ -28,6 +29,10 @@ class ActionClient extends EventEmitter {
 
     if (GoalID === null) {
       GoalID = msgUtils.requireMsgPackage('actionlib_msgs').msg.GoalID;
+    }
+
+    if (Header === null) {
+      Header = msgUtils.requireMsgPackage('std_msgs').msg.Header;
     }
 
     this._actionType = options.type;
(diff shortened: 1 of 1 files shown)


chfritz:
:+1:
@chris-smith, @IanTheEngineer : FYI, @mjyc is a member of my team. We are starting to test rosnodejs in production, so we will probably be finding and fixing a bunch of smaller issues over the next couple of months or so. It would be great if we could get those merged. But please let me know if you would prefer us collecting them in my fork for now, so we can get those fixes out quickly, and only merge them into kinetic-devel later in one batch.

src/lib/ActionClient.js
```diff
@@ -30,6 +31,10 @@ class ActionClient extends EventEmitter {
       GoalID = msgUtils.requireMsgPackage('actionlib_msgs').msg.GoalID;
     }
 
+    if (GoalID === null) {
```
chris-smith: `if (Header === null) { ` ?

chris-smith:
Awesome! We should definitely be able to merge things in - response times may vary depending on our schedule but I wouldn't expect anything to linger too long.

src/lib/ActionClient.js
IanTheEngineer: @mjyc I agree with @chris-smith here. Looks like you're intending to check and then set the `Header`. I'd be happy to merge after the change.

IanTheEngineer:
@chfritz I'm thrilled to hear that your crew is starting to use `rosnodejs` in production, and certainly welcome any / all fixes that you uncover along the way. Both Chris and I will attempt to keep on top of issues & PR.

chfritz:
Awesome, thanks @IanTheEngineer & @chris-smith !

src/lib/ActionClient.js
mjyc: @IanTheEngineer @chris-smith nice to meet you. I fixed the L34 as suggested.

IanTheEngineer:
Great! Thanks for the updated commit.

PR #51, opened 2017-02-23
Multiple pub subs

chris-smith:
- Allows multiple independent publishers/subscribers on the same topic so shutting down one will not shut down the others.
- Adds support for the `getBusInfo` XMLRPC call (this is what `rosnode info` uses)
- Adds partial support for a maximum number of XMLRPC calls before failure. Expanding this will let users
  attempt to connect to a ROS master and handle the failure cleanly on their end if there isn't one there.
- No longer supports multiple calls to `advertiseService` (mirroring `roscpp`)

`PublisherImpl` and `SubscriberImpl` are essentially moves of the previous Publisher, Subscriber code.

### DIFF
--- a/src/lib/Subscriber.js
+++ b/src/lib/Subscriber.js
@@ -64,7 +64,7 @@ class Subscriber extends EventEmitter {
   }
 
   /**
-   * Get the numbber of publishers currently connected to this subscriber
+   * Get the number of publishers currently connected to this subscriber
    * @returns {number}
    */
   getNumPublishers() {
--- a/src/lib/Subscriber.js
+++ b/src/lib/Subscriber.js
@@ -18,6 +18,7 @@
 "use strict";
 
 const EventEmitter = require('events');
+const Ultron = require('ultron');
 const {rebroadcast} = require('../utils/event_utils.js');
 
 //-----------------------------------------------------------------------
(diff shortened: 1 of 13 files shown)


src/lib/Subscriber.js
```diff
@@ -17,294 +17,89 @@
...
-  }
-
-  isShutdown() {
-    return this._state === SHUTDOWN;
-  }
-
   /**
-   * Send a topic request to each of the publishers we haven't connected to yet
-   * @param pubs {Array} array of uris of nodes that are publishing this topic
+   * Get the numbber of publishers currently connected to this subscriber
```
IanTheEngineer: Get the *number*

src/lib/impl/PublisherImpl.js
```diff
@@ -0,0 +1,330 @@
+/*
+ *    Copyright 2016 Rethink Robotics
```
IanTheEngineer: 2017 copyrights for new files

IanTheEngineer (approved):
Looks great! Thank you for splitting Pubs & Subs out into impl's like the `roscpp`  and `rospy` implementations. In addition to allowing multiple clients to open/close on the same topic, I think this will make things more maintainable for the future.

rudicus (approved):

IanTheEngineer (approved):
LGTM :ship:

PR #52, opened 2017-03-15
Add .npmignore

chris-smith:
Patch version bump


IanTheEngineer:
Looks good to me. Just for posterity, this is meant to fix the issue where NPM is using the `.gitignore` by default as the `.npmignore`. This results in only being able to use `rosnodejs` with `nodejs 6.x` or higher. The fix will allow `rosnodejs` to be used with `nodejs 4.x` and higher.

PR #58, opened 2017-05-22
Add options for action client

chris-smith:
Allows users to specify overrides for action client publisher and subscriber default settings.

### DIFF
--- a/src/lib/NodeHandle.js
+++ b/src/lib/NodeHandle.js
@@ -185,7 +185,7 @@ class NodeHandle {
     }
   }
 
-  actionClient(actionServer, type) {
+  actionClient(actionServer, type, options={}) {
     if (!actionServer) {
       throw new Error(`Unable to create action client to unspecified server - [${actionServer}]`);
     }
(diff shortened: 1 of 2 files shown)


chris-smith:
@RethinkRobotics-opensource/rethink-engine - can someone look at this quickly? It is identical to my pull-request [here](https://github.com/chris-smith/rosjs/pull/9) that I then forgot to pull in. This is necessary before I can update our `rosnodejs` version.

src/lib/NodeHandle.js
```diff
@@ -195,11 +195,11 @@ class NodeHandle {
 
     // don't namespace action client - topics will be resolved by
     // advertising through this NodeHandle
-    return new ActionClient({
+    return new ActionClient(Object.assign(options, {
```
rethink-rlinsalata: Rather than editing the user's referenced `options` object, should use a dummy empty object as the target:
`Object.assign({}, options, {actionServer, ...})`

rethink-rlinsalata (changes requested):
One fix/comment on potential side-effects. Besides that looks good!

src/lib/NodeHandle.js
chris-smith: good call

rethink-rlinsalata (approved):

PR #61, opened 2017-07-07
Spinner and lifecycle

chris-smith:
-Users can now configure default spinner behavior or provide their own.
-Provides future support for us to create other spinners that users
 can choose to deploy.
-Publishers and Subscribers with queue sizes < 1 will now have
 infinite queues. Previously you needed to disable throttling
 to get this behavior.
-Fix bugs in shutdown methods.
-Allow users to hook into shutdown event
 and query for isShutdown.
-Thread options through master API client.
-Add shutdown tests.


rethink-rlinsalata (approved):
Looks A+

PR #62, opened 2017-07-11
Subscriber shutdown bug

chris-smith:
Fixes an issue where publisher clients in the process
of being validated will not be destroyed when the
subscriber is shutdown.

### DIFF
--- a/src/lib/impl/SubscriberImpl.js
+++ b/src/lib/impl/SubscriberImpl.js
@@ -78,6 +78,8 @@ class SubscriberImpl extends EventEmitter {
 
     this._pubClients = {};
 
+    this._pendingPubClients = {};
+
     this._state = REGISTERING;
     this._register();
   }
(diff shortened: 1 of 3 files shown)


src/lib/impl/SubscriberImpl.js
```diff
@@ -131,10 +133,12 @@ class SubscriberImpl extends EventEmitter {
     this._log.debug('Shutting down subscriber %s', this.getTopic());
 
     Object.keys(this._pubClients).forEach(this._disconnectClient.bind(this));
+    Object.keys(this._pendingPubClients).forEach(this._disconnectClient.bind(this));
```
rethink-rlinsalata, line 136: This won't really do anything though, because the clients won't be in _pubClients though, right?

src/lib/impl/SubscriberImpl.js
```diff
@@ -212,7 +216,7 @@ class SubscriberImpl extends EventEmitter {
   _disconnectClient(clientId) {
     const client = this._pubClients[clientId];
     if (client) {
```
rethink-rlinsalata: Shouldn't this have an else statement that checks for `this._pendingPubClients[clientId]` ?

rethink-rlinsalata (changes requested):

src/lib/impl/SubscriberImpl.js
chris-smith: It definitely should.

src/lib/impl/SubscriberImpl.js
chris-smith: Yuuup. Now resolved.

rethink-rlinsalata (approved):

PR #64, opened 2017-08-04
Fix issue with service client call queuing

chris-smith:
Check if there's a call in progress when deciding if we should actually start a new call.

### DIFF
--- a/src/lib/ServiceClient.js
+++ b/src/lib/ServiceClient.js
@@ -137,9 +137,9 @@ class ServiceClient extends EventEmitter {
         oldCall.reject(err);
       }
 
-      // if there weren't any other calls in the queue, execute this new call
+      // if there weren't any other calls in the queue and there's no current call, execute this new call
       // otherwise new call will be handled in order when others complete
-      if (this._callQueue.length === 1) {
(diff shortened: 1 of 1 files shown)


src/lib/ServiceClient.js
```diff
@@ -137,9 +137,9 @@ class ServiceClient extends EventEmitter {
         oldCall.reject(err);
       }
 
-      // if there weren't any other calls in the queue, execute this new call
+      // if there weren't any other calls in the queue and there's no current call, execute this new call
       // otherwise new call will be handled in order when others complete
-      if (this._callQueue.length === 1) {
+      if (this._callQueue.length === 1 && this._currentCall === null) {
```
rethink-rlinsalata, line 142: should you check if you're shutdown or the currentCall is in any other sort of error state?

rethink-rlinsalata (approved):

src/lib/ServiceClient.js
chris-smith: I think we're ok here - `_executeCall` checks if we're shutdown and any call ending (successful or otherwise) should trigger the client to see if anything is queued and execute that call.

IanTheEngineer (approved):
LGTM

PR #65, opened 2017-08-04
Action Server Support

chris-smith:
-Adds interface for ActionServers
-Adds experimental ActionServer implementation. I
 assume there are still bugs here. Protocol generally
 unspecified - had to hunt through actionlib source code.
-Renames ActionClient to ActionClientInterface.
-Adds time utilities.
-Throw useful errors when invalid message type was specified.

### DIFF
--- a/src/actions/ActionServer.js
+++ b/src/actions/ActionServer.js
@@ -1,0 +1,228 @@
+/*
+ *    Copyright 2017 Rethink Robotics
+ *
+ *    Copyright 2017 Chris Smith
+ *
+ *    Licensed under the Apache License, Version 2.0 (the "License");
+ *    you may not use this file except in compliance with the License.
+ *    You may obtain a copy of the License at
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *    Unless required by applicable law or agreed to in writing, software
+ *    distributed under the License is distributed on an "AS IS" BASIS,
+ *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ */
+
+'use strict';
+
+const timeUtils = require('../lib/Time.js');
+const msgUtils = require('../utils/message_utils.js');
+const EventEmitter = require('events');
+
+const ActionServerInterface = require('../lib/ActionServerInterface.js');
+const GoalHandle = require('./GoalHandle.js');
+
+let GoalIdMsg = null;
+let GoalStatusMsg = null;
+let GoalStatusArrayMsg = null;
+let GoalStatuses = null;
+let goalCount = 0;
+
+/**
+ * @class ActionServer
+ * EXPERIMENTAL
+ *
+ */
+class ActionServer extends EventEmitter {
+  constructor(options) {
+    super();
+
+    if (GoalStatusMsg === null) {
+      GoalStatusMsg = msgUtils.requireMsgPackage('actionlib_msgs').msg.GoalStatus;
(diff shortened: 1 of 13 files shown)


src/actions/ActionServer.js
```diff
@@ -0,0 +1,230 @@
...
+  _getGoalHandle(id) {
+    return this._goalHandleCache[id];
+  }
+
+  _handleGoal(msg) {
+    const newGoalId = msg.goal_id.id;
+
+    let handle = this._getGoalHandle(newGoalId);
+
+    let AddedGoal
```
IanTheEngineer: I don't see this variable being used elsewhere. Should it be removed?

src/actions/GoalHandle.js
```diff
@@ -0,0 +1,182 @@
...
+      case GoalStatuses.RECALLING:
+        this._setStatus(GoalStatuses.RECALLED, text);
+        this._publishResult(result);
+        break;
+      case GoalStatuses.PREEMPTING:
+        this._setStatus(GoalStatuses.PREEMPTED, text);
+        this._publishResult(result);
+        break;
+      default:
+        // FIXME: log errors for invalid state transitions
```
IanTheEngineer: As a reference, perhaps toss in this link:
http://wiki.ros.org/actionlib/DetailedDescription#Server_Description
to aid the future fixer :)

IanTheEngineer: In the mean time, just logging any error about an unknown transition would be helpful...

src/utils/XmlrpcClient.js
```diff
@@ -22,10 +22,11 @@ class XmlrpcCall {
           reject(err);
         }
         else if (resp[0] !== 1) {
-          const error = new Error('ROS XMLRPC Error');
+          const msg = resp[1];
+          const error = new Error(`ROS XMLRPC Error: ${msg}`);
```
IanTheEngineer, line 26: Is this fixing a separate bug?

src/utils/time_utils.js
```diff
@@ -17,21 +17,48 @@
 
 'use strict';
 
+const NSEC_TO_SEC = 1e-9;
```
IanTheEngineer, line 20: +1 for actual constants!

IanTheEngineer (approved):
Looks good, just a few small comments & logging requests.

IanTheEngineer:
Appears this branch needs to be rebased...

src/actions/ActionServer.js
chris-smith: seems like it

src/utils/XmlrpcClient.js
chris-smith: This isn't a bug - this is a response from a remote XMLRPC server that indicates some problem in handling the request. It will happen if you try to get a parameter or lookup a service that doesn't exist.

http://wiki.ros.org/ROS/Master_Slave_APIs

src/index.js
```diff
@@ -295,9 +297,10 @@ let Rosnodejs = {
         goal: { edges: 3,  radius: 1 } }));
    */
   getActionClient(options) {
-    options.nh = this.nh;
-    return new ActionClient(options);
+    return this.nh.actionClientInterface(options);
```
chfritz, line 300: This doesn't actually work. The [invoked function](https://github.com/RethinkRobotics-opensource/rosnodejs/blob/e68811c43a345782da863b7bd7c66498ad6d9729/src/lib/NodeHandle.js#L200) doesn't take an object as argument, it requires the server and type to be broken out. I'll create a PR to fix that.

PR #66, opened 2017-08-22
Fixes a typo in on-the-fly message generation code

chfritz:
Small but catastrophic.

Without it, the initNode fails with
```
[ERROR] [1503360005.360] (ros.rosnodejs): Exception trying to advertise service /app/get_loggers
```


chris-smith:
Yikes! Thanks for catching that

PR #68, opened 2017-09-06
Support asynchronous service callbacks

chris-smith:
-Service callbacks can now return promises. Result
 will be sent when promise resolves.
-Adds tests for asynchronous services.
-Service clients whose connections are closed while a call
 is in progress will be rejected.

### DIFF
--- a/src/lib/ServiceClient.js
+++ b/src/lib/ServiceClient.js
@@ -235,7 +235,14 @@ class ServiceClient extends EventEmitter {
     call.serviceClient.write(serializedRequest);
 
     return new Promise((resolve, reject) => {
+      const closeHandler = () => {
+        this._log.debug('Service %s client disconnected during call!', this.getService());
+        reject(new Error('Connection was closed'));
+      }
+
       call.serviceClient.$deserializeStream.once('message', (msg, success) => {
+        call.serviceClient.removeListener('close', closeHandler);
+
         if (success) {
           resolve(this._messageHandler.Response.deserialize(msg));
         }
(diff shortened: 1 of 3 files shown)


IanTheEngineer (approved):
Looks great! I think this should be good to merge after a rebase

blakelyc:
Correct me if I am wrong, but this why Actions exist.  Services are supposed to be synchronous (blocking).

chris-smith:
TLDR: I think enforcing a synchronous Service callback pattern breaks down a bit in Node.

In this patch we're not stopping people from writing simple, synchronous callback functions that return immediately - if their callback function returns a boolean response we'll consider the service complete. I do think it'd be strange if users returned `true` in their callback but also kicked off some asynchronous work.

But if in order to do the service's job it has to do some asynchronous work (which is very common in Node) then I think we should let them and should delay sending the response until they're done. These services don't necessarily need to support preemption or provide progress, but they may still need to be asynchronous because of the nature of the language and the context in which the service is being provided.

Blocking service callbacks in Node.js will also forcibly block anything else from happening (subscription callbacks, web server responses, etc.) which isn't necessarily the case in other implementations (`roscpp` or `rospy`) where that work can be done in separate threads. Actions may be a bit heavy-handed for a lot of these services - I'm assuming actions require more overhead as well because of their complexity.

I'd also take the position that Actions exist to support multiple long-running goals, provide progress on those goals, and allow them to be canceled or interrupted cleanly while Services exist to provide a simple call-and-response mechanism. I've always assumed that the synchronous-Service-callback-pattern arose because callback function signatures implied so in dominant implementations - and because actions came along that made it unnecessary to implement services any differently for those languages.

All that being said, I'm open to discussing this with other people in the Node/ROS community.

NickZ:
Synchronous service callbacks are the bane of my existence, and have caused us so many problems in trying to support services in our software. I would love love love for this to be merged in.

chris-smith:
Barring some negative response to this, I'll likely merge in the next couple of days.

PR #81, opened 2017-12-08
Implement timeout on initNode()

NickZ:
This implements a timeout option for initNode. initNode will timeout with a rejection if the node is not initialized within the specified number of seconds. If the timeout is negative, then it will keep retrying forever.

Also fixed some of the Promise logic in InitNode().

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -125,8 +125,9 @@ let Rosnodejs = {
    * @param {function}  options.logging.setLoggerLevel  the function for setting the logger
    *                                                    level
    * @param {string}    options.rosMasterUri the Master URI to use for this node
-   * @param {number}    options.timeout time in seconds to wait for node to be initialized
-   *                                    before timing out
+   * @param {number}    options.timeout time in ms to wait for node to be initialized
+   *                                    before timing out. A negative value will retry forever.
+   *                                    A value of '0' will try once before stopping. @default -1
    * @return {Promise} resolved when connection to master is established
    */
   initNode(nodeName, options) {
--- a/src/index.js
+++ b/src/index.js
@@ -163,7 +163,7 @@ let Rosnodejs = {
       this._loadOnTheFlyMessages(options)
       .then(()=>{return _checkMasterHelper(100, options.timeout);})
       .then(Logging.initializeRosOptions.bind(Logging, this, options.logging))
-      .then(Time._initializeRosTime.bind(Time, this))
+      .then(Time._initializeRosTime.bind(Time, this, options.notime))
       .then(() => { resolve(this.getNodeHandle()); })
       .catch((err) => {
         log.error('Error during initialization: ' + err);
--- a/src/index.js
+++ b/src/index.js
@@ -48,19 +48,18 @@ let pingMasterTimeout = null;
  * Helper function to see if the master is available and able to accept
  * connections.
  * @param {number} timeout time in ms between connection attempts
- * @param {number} maxTimeout maximum time in seconds to retry before timing out. 
+ * @param {number} maxTimeout maximum time in ms to retry before timing out. 
  * A negative number will make it retry forever. 0 will only make one attempt
  * before timing out.
  */
 function _checkMasterHelper(timeout=100, maxTimeout=-1) {
-  let startTime = Time.now().secs;
+  let startTime = Date.now();
   const localHelper = (resolve,reject) => {
     pingMasterTimeout = setTimeout(() => {
       // also check that the slave api server is set up   
       if (!rosNode.slaveApiSetupComplete()) {
-        console.error("hello slave api isnt setup yet");
-        if (Time.now().secs - startTime >= maxTimeout && !(maxTimeout < 0) ) {
-          let error = `Unable to register with master node [${rosNode.getRosMasterUri()}]: unable to set up slave API Server. Stopping...`
+        if (Date.now() - startTime >= maxTimeout && !(maxTimeout < 0) ) {
+          log.error(`Unable to register with master node [${rosNode.getRosMasterUri()}]: unable to set up slave API Server. Stopping...`);
           reject(Error('Unable to setup slave API server.'));
           return;
         }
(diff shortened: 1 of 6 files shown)


src/index.js
```diff
@@ -113,8 +146,8 @@ let Rosnodejs = {
         return Promise.resolve(this.getNodeHandle());
       }
       // else
-      throw new Error('Unable to initialize node [' + nodeName + '] - node ['
-                      + rosNode.getNodeName() + '] already exists');
+      return Promise.reject( Error('Unable to initialize node [' + nodeName + '] - node ['
```
chris-smith, line 146: My main concern with this change is that it changes existing behavior.

src/index.js
```diff
@@ -129,14 +162,18 @@ let Rosnodejs = {
...
+    return new Promise((resolve,reject)=>{
+      this._loadOnTheFlyMessages(options)
+      .then(()=>{return _checkMasterHelper(100, options.timeout);})
       .then(Logging.initializeRosOptions.bind(Logging, this, options.logging))
       .then(Time._initializeRosTime.bind(Time, this))
-      .then(() => { return this.getNodeHandle(); })
+      .then(() => { resolve(this.getNodeHandle()); })
       .catch((err) => {
         log.error('Error during initialization: ' + err);
+        this.shutdown();
```
chris-smith: this might throw additional errors depending on when the error occurred - the shutdown call in the RosNode assumes the initialization happened appropriately

src/index.js
```diff
@@ -129,14 +162,18 @@ let Rosnodejs = {
     const nodeOpts = options.node || {};
     rosNode = new RosNode(nodeName, rosMasterUri, nodeOpts);
 
-    return this._loadOnTheFlyMessages(options)
-      .then(_checkMasterHelper)
+    return new Promise((resolve,reject)=>{
```
chris-smith, line 162: does explicitly creating a new promise here do something that I'm missing?

src/index.js
```diff
@@ -43,13 +43,28 @@ let rosNode = null;
...
+ * @private
+ * Helper function to see if the master is available and able to accept
+ * connections.
+ * @param {number} timeout time in ms between connection attempts
+ * @param {number} maxTimeout maximum time in seconds to retry before timing out. 
+ * A negative number will make it retry forever. 0 will only make one attempt
+ * before timing out.
+ */
+function _checkMasterHelper(timeout=100, maxTimeout=-1) {
+  let startTime = Time.now().secs;
```
chris-smith: I'd lean towards using Date.now() for these instead - Time isn't initialized until the Ros node is initialized. Worth adding a comment to the effect too.

src/index.js
```diff
@@ -43,13 +43,28 @@ let rosNode = null;
...
 //------------------------------------------------------------------
-
-function _checkMasterHelper(timeout=100) {
-  const localHelper = (resolve) => {
+/**
+ * @private
+ * Helper function to see if the master is available and able to accept
+ * connections.
+ * @param {number} timeout time in ms between connection attempts
+ * @param {number} maxTimeout maximum time in seconds to retry before timing out.
```
chris-smith: `rosnodejs` generally uses ms for time units - is there a reason you decided to go with seconds?

src/index.js
```diff
@@ -43,13 +43,28 @@ let rosNode = null;
...
+ */
+function _checkMasterHelper(timeout=100, maxTimeout=-1) {
+  let startTime = Time.now().secs;
+  const localHelper = (resolve,reject) => {
     pingMasterTimeout = setTimeout(() => {
-      // also check that the slave api server is set up
+      // also check that the slave api server is set up   
       if (!rosNode.slaveApiSetupComplete()) {
-        localHelper(resolve);
+        console.error("hello slave api isnt setup yet");
```
chris-smith: debug log

src/index.js
```diff
@@ -43,13 +43,28 @@ let rosNode = null;
...
+        if (Time.now().secs - startTime >= maxTimeout && !(maxTimeout < 0) ) {
+          let error = `Unable to register with master node [${rosNode.getRosMasterUri()}]: unable to set up slave API Server. Stopping...`
```
chris-smith: did you mean to do something with this variable? or am I missing where it's being used

src/index.js
```diff
@@ -59,14 +74,23 @@ function _checkMasterHelper(timeout=100) {
         resolve();
       })
       .catch((err, resp) => {
-        log.warnThrottle(60000, `Unable to register with master node [${rosNode.getRosMasterUri()}]: master may not be running yet. Will keep trying.`);
-        localHelper(resolve);
+        let error = `Unable to register with master node [${rosNode.getRosMasterUri()}]: master may not be running yet.`
+        if (Time.now().secs - startTime >= maxTimeout && !(maxTimeout < 0) ){
+          error += ' Timed out! Stopping...';
+          log.error(error);
```
chris-smith: can you log a separate message here rather than manipulating the 'base' error?
`Timed out before registering with master node [${rosNode.getRosMasterUri()}]: master may not be running yet.`

chris-smith (changes requested):
Awesome! a couple of small comments.

src/index.js
NickZ: Since this was already returning a promise and the user was expecting that the promise would either resolve or reject (and set up the appropriate `.catch()`), I'm of the opinion that this should just reject instead of throwing an error, unless you want this to be explicitly fatal.

src/index.js
NickZ: We were previously not rejecting when an error occurred during initialization (and my timeout logic wouldn't work), I added this so I could add an explicit rejection in the catch block below.

src/index.js
NickZ: I think that's fine; the exit implementation in rosnode catches all the potential rejections that could happen while shutting down, the only thing that would happen is some extra errors in the log.

src/index.js
chris-smith: That was my original thought - in my mind this was more explicit user-error. The distinction seems less noticeable in newer versions of JS though.

src/index.js
chris-smith: The errors I was worried about don't seem to happen - but this line (and rejecting with the error below) do break a lot of the tests.

chris-smith (commented):
mostly concerned about the broken tests and the implication for users. I think we'd need a major version change to release this.

NickZ:
I fixed the issues with the tests; Previously, the tests were initializing a test node on a rump xmlrpc server; this server did not have a parameter server, and so `initNode()` would experience errors during initialization (specifically during initialize Time). However, since the node was never shut down, and `initNode()` never rejected, the tests could keep using the test node just fine. So, what I did was I added a "notime" option to `initNode()` which skips the Time initialization, and now the tests work with my added changes.

I agree this should probably wait for a major version since this could break users' programs, but this behavior probably should not have existed in the first place; if the node experiences errors during initialization, we should at least reject with an error so that the user doesn't assume that the node is OK to use.

NickZ:
Thank you for merging this in!

BTW, you are doing god's work by developing rosnodejs. It's definitely become a big part of our software strategy.

PR #89, opened 2018-03-21
Add Initial Remapping Support

### LINKED ISSUE #82: Remapping is not working
Hi.

I am trying to use rosnodejs for my project, but it is ignoring the remappings set by the command-line arguments or roslaunch files.

I have a node like this
`rosnodejs.initNode('foo').then(....);`
and want to rename it with roslauch like this
`<node pkg="..." name="foo" type="...">
  <remap from="foo" to="bar"/>
</node>`
, but `rosnode list` is just showing /foo.

Is there a quick way to fix this problem?

I am using the version 2.1.1.

Thank you.

chris-smith:
Resolves #82
-Adds the ability to remap topics, services, and parameters
 through the command line. Also supports special remapping keys
 for node name, ip, host, master and namespace.
-Adds tests for remapping
-Adds tests for some new node initialization options.


louisgv:
@chris-smith is there anything I can help to get this PR merged? Seems like it was neglected for quite a while.

chris-smith:
I mostly meant to push someone to review it since it touches very public code.

src/index.js
```diff
@@ -36,81 +36,15 @@ const packages = require('./utils/messageGeneration/packages.js');
...
- * Appends a random string of numeric characters to the end
- * of the node name. Follows rospy logic.
- * @param nodeName {string} string to anonymize
- * @return {string} anonymized nodeName
- */
-function _anonymizeNodeName(nodeName) {
-  return util.format('%s_%s_%s', nodeName, process.pid, Date.now());
-}
 
 let Rosnodejs = {
```
louisgv, line 49: If this reference is not mutated later, should use `const` instead.

src/index.js
```diff
@@ -338,3 +282,84 @@ let Rosnodejs = {
...
+ * @private
+ * Helper function to see if the master is available and able to accept
+ * connections.
+ * @param {number} timeout time in ms between connection attempts
+ * @param {number} maxTimeout maximum time in ms to retry before timing out.
+ * A negative number will make it retry forever. 0 will only make one attempt
+ * before timing out.
+ */
+function _checkMasterHelper(timeout=100, maxTimeout=-1) {
+  let startTime = Date.now();
```
louisgv: Did not seem to mutate, should use `const`

src/index.js
```diff
@@ -338,3 +282,84 @@ let Rosnodejs = {
...
+  const localHelper = (resolve,reject) => {
+    pingMasterTimeout = setTimeout(() => {
+      // also check that the slave api server is set up
+      if (!rosNode.slaveApiSetupComplete()) {
+        if (Date.now() - startTime >= maxTimeout && !(maxTimeout < 0) ) {
```
louisgv: Should prefer simple condition. I.e `maxTimeout >= 0`

src/index.js
```diff
@@ -338,3 +282,84 @@ let Rosnodejs = {
...
+  return new Promise((resolve, reject) => {
+    localHelper(resolve,reject);
+  });
+}
+
+function _resolveNodeName(nodeName, remappings, options) {
+  let namespace = remappings['__ns'] || process.env.ROS_NAMESPACE || '';
+  namespace = names.clean(namespace);
+  if (namespace.length === 0 || !namespace.startsWith('/')) {
+    namespace = '/' + namespace;
```
louisgv: Be consistent with string literal usage. E.g since template is widely used in this repo, should use:
```
namespace = `/${namespace}`
```

src/lib/Names.js
```diff
@@ -0,0 +1,211 @@
...
+  else if (code >= 48 && code <= 57) {
+    return true;
+  }
+  // else
+  return false;
+}
+
+function isValidCharInName(char) {
+  if (isAlnum(char) || char == '/' || char == '_') {
+    return true;
```
louisgv: This code seems overly clear. The minor suggestion is to remove the if-else and return the condition immediately. i.e

```
return (isAlnum(char) || char == '/' || char == '_')
```

The major suggestion is to use strict comparison: 

```
isAlnum(char) || char === '/' || char === '_')
```

src/lib/Names.js
```diff
@@ -0,0 +1,211 @@
...
+
+//------------------------------------------------------------------
+// Local Helper functions
+//------------------------------------------------------------------
+
+
+
+function isAlpha(char, code = null) {
+  code = code || char.charCodeAt(0);
+  if (code >= 65 && code <= 90) {
```
louisgv: JS char comparison already uses charCode internally so technically we can do it like this:
```
char >= 'A' && char <= 'Z'
```

I did a jsperf and it's faster than do charCodeAt:

image

Perf at: https://jsperf.com/char-code-comparison

src/index.js
```diff
@@ -338,3 +282,84 @@ let Rosnodejs = {
...
+ */
+function _checkMasterHelper(timeout=100, maxTimeout=-1) {
+  let startTime = Date.now();
+  const localHelper = (resolve,reject) => {
+    pingMasterTimeout = setTimeout(() => {
+      // also check that the slave api server is set up
+      if (!rosNode.slaveApiSetupComplete()) {
+        if (Date.now() - startTime >= maxTimeout && !(maxTimeout < 0) ) {
+          log.error(`Unable to register with master node [${rosNode.getRosMasterUri()}]: unable to set up slave API Server. Stopping...`);
+          reject(Error('Unable to setup slave API server.'));
```
louisgv: Might want to be consistent and use new Error.

src/lib/Names.js
```diff
@@ -0,0 +1,211 @@
...
+      default:
+        [namespace, name, remap] = args;
+        break;
+    }
+
+    return [namespace, name, remap];
+  }
+}
+
+module.exports = new Names();
```
louisgv, line 173: Exporting an instance to mock the behavior of a singleton is usually bad practice in my opinion. It is better to create a singleton that is named something like `Store` which instantiate and export singleton instance for other module to consume.

src/lib/Names.js
```diff
@@ -0,0 +1,211 @@
...
+    const c = name[0];
+    if (!isAlpha(c) && c !== '/' && c !== '~') {
+      if (throwError) {
+        throw new Error(`Character [${c}] is not valid as the first character in Graph Resource Name [${name}].  Valid characters are a-z, A-Z, / and in some cases ~.`);
+      }
+      // else
+      return false;
+    }
+
+    for (let i = 1; i < len; ++i) {
```
louisgv, line 49: RegExp test might be better for this based on the perf below:

image

Just replace the first character with `-` maybe then the regexp should work.

DYI: https://jsperf.com/alphanumeric-string-test

src/utils/network_utils.js
```diff
@@ -17,13 +17,23 @@
...
-let os = require('os');
+const os = require('os');
+const RemapUtils = require('./remapping_utils');
 
-const ROS_IP = process.env.ROS_IP;
-const ROS_HOSTNAME = process.env.ROS_HOSTNAME;
-const HOST = ROS_IP || ROS_HOSTNAME || os.hostname();
+let HOST = null;
 
 let NetworkUtils = {
```
louisgv: NetworkUtils should be const as it is not mutated.

test/namespaceTest.js
```diff
@@ -3,59 +3,352 @@
...
   });
-});
 
+  describe('with ros', () => {
+    const MASTER_PORT = 12342;
+    const rosMasterUri = `http://localhost:${MASTER_PORT}`;
+    let masterStub = new MasterStub('localhost', MASTER_PORT);
+    masterStub.provideAll();
+
+    let ARGV_LEN = process.argv.length;
```
louisgv: Should be const

louisgv:
@chris-smith I made some review and do some unit perf on some of the change. I will look into running the test. Does this work with just the turtle sim?

PR #90, opened 2018-03-21
Update .travis.yml

chris-smith:
-Add minimal ros install in order to actually run tests.
-Currently just runs against 14.04 and indigo. Also running against 16.04 and kinetic will require setting up travis with docker.
-compiles, generates messages, and tests against latest of Node 6, 7, and 8.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,69 @@
+dist: trusty
+sudo: required
+compiler:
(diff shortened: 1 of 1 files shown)


IanTheEngineer (approved):

rethink-neil:
+1

PR #91, opened 2018-03-21
Fixed how GoalHandle's status field is accessed in ActionServer.

jstnhuang:
The status field of a GoalHandle is this._status, not this.status. I updated ActionServer to match.

PR #92, opened 2018-03-21
Added goal to GoalHandle

### LINKED ISSUE #88: Contents of action goal not exposed?
If we want to write an action server, how do we get the contents of the goal message? It looks like when an ActionServer constructs a GoalHandle, it only provides a goal ID, but the contents of the goal are not saved anywhere.

jstnhuang:
Relevant to #88.

This change makes the ActionServer pass the goal message to a GoalHandle, which is stored in the `.goal` field. When GoalHandles are created for cancellations, the goal message is left undefined.

This incorporates changes from #91. I put the changes from #91 into a separate pull request since it's a straightforward typo fix, whereas I'm not sure if this is the best way to fix #88.

### DIFF
--- a/src/actions/ActionServer.js
+++ b/src/actions/ActionServer.js
@@ -103,7 +103,7 @@ class ActionServer extends EventEmitter {
       return false;
     }
 
-    handle = new GoalHandle(msg.goal_id, this);
+    handle = new GoalHandle(msg.goal_id, this, GoalStatuses.PENDING, msg.goal);
     this._goalHandleList.push(handle);
     this._goalHandleCache[handle.id] = handle;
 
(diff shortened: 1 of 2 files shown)


IanTheEngineer (approved):

IanTheEngineer:
Looks really good! Thanks for the addition Justin. You'll need to rebase onto the current `kinetic-devel` branch before we can merge due to the conflict with you previous PR.

chris-smith:
It may make sense for goal to be "private" (`GoalHandle._goal`) and accessable through a `getGoal()` function to mirror what [GoalHandle](http://docs.ros.org/jade/api/actionlib/html/classactionlib_1_1ServerGoalHandle.html#a5eef2ff08ae8ca77682c39fadb09ff11) in actionlib does.

Otherwise this all seems like the right thing to do.

jstnhuang:
Oops, I forgot I had clang-format on for JavaScript. I'll let you know when I'm ready for another review.

jstnhuang:
Okay, ready.

chris-smith (approved):

chris-smith:
Awesome! Thanks!

Issue #93, opened 2018-03-23
Custom messages not found?


renatosamperio:
Hi,

Many thanks for your library as it has become very useful. However, recently I had found some issues that perhaps you could help me to clarify.

I am trying to make use of custom defined messages from another module but I receive the error:
`TypeError: Cannot read property 'msg' of undefined`

I am including the custom message module in the **package.xml** as `build_depend` and `exec_depend` and in the **CMakelists.txt** in the parts related to `find_package` and `CATKIN_DEPENDS` of the "catkin_package".

I also tried to load the messages explicitly with `rosnodejs.loadAllPackages();` before calling the `require` of the **custom-made module**, but this only works when I execute the node inside the ROS module.

I installed the **node_modules/** inside the ROS module itself. The file structure looks like this:

>my_ws/
>> custom_messages_module/
>>> msg/
>>>> myMessage.msg

>> error_module/
>>> scripts/
>>>>> node_modules/
>>>>> package-lock.json
>>>>> ros_node_script.js

`user@host: ~/my_ws$ rosrun  error_module ros_node_script.js`
`TypeError: Cannot read property 'msg' of undefined`

The file **ros_node_script.js** would look like this:
`#!/usr/bin/env node`
`const rosnodejs = require('rosnodejs');`
`rosnodejs.loadAllPackages();`
`const custom_messages_module = rosnodejs.require('custom_messages_module').msg;`

Any idea would be very helpful.

Regards,
Renato.

chris-smith:
`rosnodejs.loadAllPackages` is asynchronous and so it won't have managed to build anything by the time you try to require them. Waiting for the promise to resolve would likely not resolve the issue either though as we only search for messages on disk once when `rosnodejs` is first loaded.

`loadAllPackages` is meant as a one-time build step, so I'd recommend running it on its own and verifying that it mentions your custom_messages_module somewhere in the output

```
$ node
> const rosnodejs = require('rosnodejs');
> rosnodejs.loadAllPackages();
Promise { <pending> }
> Finished building package topic_tools
Finished building package std_msgs
Finished building package nodelet
Finished building package std_srvs
Finished building package custom_messages_module
Finished building messages!
```
After this, run your `ros_node_script.js` file without the `loadAllPackages()` line. You could also generate messages on the fly rather than relying on them from disk, but you wouldn't be able to require any message packages until the node has initialized
```
rosnodejs.initNode('my_node', { onTheFly: true })
.then(() => {
[...]

Issue #85, opened 2018-03-05
Standard messages are not built


bgromov:
I followed the instruction on [rosnodejs_examples](https://github.com/RethinkRobotics-opensource/rosnodejs_examples), but could not run neither of the nodes. The error is:

```
/home/user/rosnodejs_ws/src/rosnodejs_examples/scripts/talker.js:28
const std_msgs = rosnodejs.require('std_msgs').msg;
                                              ^

TypeError: Cannot read property 'msg' of undefined
    at Object.<anonymous> (/home/user/rosnodejs_ws/src/rosnodejs_examples/scripts/talker.js:28:47)
    at Module._compile (module.js:410:26)
    at Object.Module._extensions..js (module.js:417:10)
    at Module.load (module.js:344:32)
    at Function.Module._load (module.js:301:12)
    at Function.Module.runMain (module.js:442:10)
    at startup (node.js:136:18)
    at node.js:966:3
```

So, apparently the standard messages were not built. In the dev-space under `gennodejs` I only see the packages local to the workspace.

I tried to add `rosnodejs.loadAllPackages()` before loading `std_msgs`, but it made no difference.

My configuration is:

 - Ubuntu 16.04
 - ROS Kinetic
 - nodejs v4.2.6

Any clue what could be wrong?

chris-smith:
What are your `CMAKE_PREFIX_PATH` and `ROS_PACKAGE_PATH` environment variables?

bgromov:
Sorry for the delay.

I reduced the setup to a minimum (still doesn't work). There is just ROS Desktop Full workspace at `/opt/ros/kinetic/` and a single catkin overlay workspace at `/home/user/rosnodejs_ws/`. Here is the environment:

```shell
CMAKE_PREFIX_PATH=/home/user/rosnodejs_ws/devel:/opt/ros/kinetic
ROS_PACKAGE_PATH=/home/user/rosnodejs_ws/src/rosnodejs_examples:/opt/ros/kinetic/share
```

chris-smith:
I'm assuming this issue occurs for other message packages too (`geometry_msgs`, `tf2_msgs`, etc) in addition to `std_msgs`?

[...]

PR #94, opened 2018-03-27
Throw errors from requiring messages

chris-smith:
-If an error occurs while requiring a message package, throw it.
-Add tests for expected messages, throwing error when message doesn't exist

Relates to #93, #85

### DIFF
--- a/src/utils/message_utils.js
+++ b/src/utils/message_utils.js
@@ -170,12 +170,7 @@ let MessageUtils = {
   },
 
   loadMessagePackage(msgPackage) {
-    try {
-      messagePackageMap[msgPackage] = ros_msg_utils.Find(msgPackage);
-    }
-    catch (err) {
-      throw new Error(`Unable to include message package ${msgPackage} - ${err}`);
-    }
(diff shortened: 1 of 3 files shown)


src/utils/message_utils.js
```diff
@@ -170,12 +170,7 @@ let MessageUtils = {
   },
 
   loadMessagePackage(msgPackage) {
-    try {
-      messagePackageMap[msgPackage] = ros_msg_utils.Find(msgPackage);
-    }
-    catch (err) {
-      throw new Error(`Unable to include message package ${msgPackage} - ${err}`);
```
IanTheEngineer, line 177: Wasn't this already re-throwing an error? Is the issue that it was swallowing the original error message?

IanTheEngineer (approved):
Looks good for merging!

src/utils/message_utils.js
chris-smith: The other function was swallowing the error but this was obscuring the actual error by changing the stack, making it harder for people to debug issues when they occur.

IanTheEngineer: :+1:

PR #100, opened 2018-05-18
Used Time.now() instead of timeUtils.now() in action server/client

### LINKED ISSUE #98: Simulation time is not used in ActionClient / ActionServer
ActionClient and ActionServer use `timeUtils.now()` throughout, but `timeUtils.now()` gets the time from `Date.now()`. This does not work when running in simulation. Should these be replaced with `Time.now()`?

jstnhuang:
Fixes #98


chris-smith (approved):
Thanks!

PR #101, opened 2018-05-18
Check if .match is null instead of its length

schadocalex:
input uri received: `http://192.168.1.45 :41049/`

and check if match is null instead of its length
.match() return null not an empty array if not found
see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/match

### DIFF
--- a/src/utils/network_utils.js
+++ b/src/utils/network_utils.js
@@ -64,7 +64,7 @@ let NetworkUtils = {
   },
 
   getAddressAndPortFromUri(uriString) {
-    let regexStr = /(?:http:\/\/|rosrpc:\/\/)?([a-zA-Z\d\-.:]+)[ ]*:(\d+)/;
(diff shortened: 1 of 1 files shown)


chris-smith:
I would consider the space in there to make the URI invalid... as if a user had a typo when setting their ROS_IP or something. Is there a reason it shouldn't be?

The null check is definitely better here though and would give a more useful error for people.

schadocalex:
I looked for why my ROS_IP got a space and i find why. In order tu use rviz remotely, I run this command:
```
export ROS_IP=`hostname -I`
```
So I modified this command to remove the trailing whitespace:
```
export ROS_IP=`hostname -I | sed -e 's/[[:space:]]*$//'`
```
And it does work. 

So I modify the PR just for the `match` check.

chris-smith (approved):

chris-smith:
Thanks!

PR #105, opened 2018-06-22
House Keeping - Update Dependencies and Unit Test

louisgv:
@chris-smith 

I noticed the version of some dependencies is a bit outdated. I decided to create this minimal PR in hope to upgrade the dependency as well as Babel's presets to use the recommended package env.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,12 +1,17 @@
 {
-    "presets": [
-      [
-        "env",
-        {
-          "targets": {
-            "browsers": ["last 2 versions", "ie >= 7"]
-          }
-        }
-      ]
+  "presets": [
+    [
+      "env",
+      {
+        "targets": {
+          "node": 4
+        },
+        "exclude": [
+          "transform-es2015-classes",
+          "transform-es2015-block-scoping",
+          "transform-es2015-arrow-functions"
+        ]
+      }
     ]
-  }
+  ]
+}
--- a/.babelrc
+++ b/.babelrc
@@ -1,0 +1,12 @@
+{
+    "presets": [
+      [
+        "env",
+        {
+          "targets": {
+            "browsers": ["last 2 versions", "ie >= 7"]
+          }
+        }
+      ]
+    ]
+  }
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
     "stressTest": "mocha test/stress.js",
     "flatten": "node tools/flatten.js",
     "generate": "node dist/tools/generateMessages.js",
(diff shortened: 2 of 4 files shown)


louisgv:
Hmm, wonder why Travis is failing :\

chris-smith:
Looks like it's the major version bump for mocha

louisgv:
Derp, read through the travis log again. I just fixed the test! 

Sorry for the late reply, I've been finishing a final for an English class.

louisgv:
Finally test passed :p @chris-smith

package.json
```diff
@@ -7,12 +7,12 @@
...
   ],
   "scripts": {
-    "test": "mocha test/directory.js",
+    "test": "mocha --exit test/directory.js",
     "gennodejsTest": "mocha test/gennodejsTest.js test/onTheFlyMessages.js",
     "stressTest": "mocha test/stress.js",
     "flatten": "node tools/flatten.js",
     "generate": "node dist/tools/generateMessages.js",
-    "compile": "babel --presets es2015 src/ -d dist/",
+    "compile": "babel --presets env src/ -d dist/",
```
chris-smith: I think we'd need to add targets here too, right? The compilation is meant to happen before publishing the package to npm so that the published code is compatible with older versions of Node. My read of the doc for babel-preset-env is that this will just transpile for the current version.

louisgv: For sure, which target would you like the package to support? Can definitely add the flag, or just add a babelrc file for this :)

louisgv:
cc @chris-smith regarding the intended target to support

chris-smith:
I think we should continue to support the same versions we did with ES2015 for now - I've been thinking about re-examining how we transpile, etc and possibly which versions we support but I'm not sure this is the right time to change that.

louisgv:
@chris-smith  The config I added should works for IE8 and up, which should be equivalent to es2015

.babelrc
```diff
@@ -0,0 +1,12 @@
+{
+    "presets": [
+      [
+        "env",
+        {
+          "targets": {
+            "browsers": ["last 2 versions", "ie >= 7"]
```
chris-smith: it feels a little odd putting "browsers" here, even though it should satisfy the requirement to transpile for es2015... maybe we can just say node 4.x? I believe @chfritz set up the initial transpilation for 4.5

louisgv:
I fix the target to point to node 4. I also excluded features not in original node 4 (refers to the issue included in the commit for more information).

chris-smith (approved):
Awesome - thanks!

PR #106, opened 2018-06-25
Changed ActionClientInterface to expect action type names without extra "Action" appended

### LINKED ISSUE #104: Inaccurate message type names generated by ActionClientInterface
I think the way the type names for an ActionClient are generated is inaccurate. If you create an action client for a type `foo_msgs/Bar` (filename Bar.action), the correct message types are `foo_msgs/BarActionGoal`, `foo_msgs/BarActionFeedback`, etc.

This should be true even if the type name ends with "Action." For example, if the action type is `foo_msgs/BarAction` (filename BarAction.action), the correct message types should be `foo_msgs/BarActionActionGoal`, `foo_msgs/BarActionActionFeedback`, etc.

The ActionServerInterface correctly appends "ActionGoal", "ActionFeedback", etc. to the base name of the action type. However, ActionClientInterface only appends "Goal", "Feedback", etc. The string "Action" is appended beforehand in NodeHandle.js, but only after this check:
```js
if (!type.endsWith('Action')) {
  type += 'Action';
}
```

If you have an action whose type name ends with "Action," this leads the ActionClientInterface to generate type names like `foo_msgs/BarActionGoal` instead of `foo_msgs/BarActionActionGoal`. This leads to md5 matching errors with other nodes, since `foo_msgs/BarActionGoal` is also a valid type.

I think the fix should be simple: remove the three lines quoted above from NodeHandle.js, and change ActionClientInterface to always append "ActionGoal", "ActionFeedback", etc. to the type name.

Please let me know what you think. Is this just an edge case that should be fixed, or is it likely that lots of people depend on this behavior? If we want to fix it, I'd be happy to make the pull request.

jstnhuang:
Fixes #104 

Changes the ActionClientInterface such that the type name is always of the form: `my_msgs/DoSomething`, where the action was defined in a file `DoSomething.action` in the package `my_msgs`.

**This is a breaking change!**
- If you previously provided the name with the word "Action" at the end, this will no longer work:
  ```diff
  - const ac = nh.actionClientInterface('/turtle_shape', 'turtle_actionlib/ShapeAction');
  ```
  You must remove the extra "Action" from the action type name:
  ```diff
  + const ac = nh.actionClientInterface('/turtle_shape', 'turtle_actionlib/Shape');
  ```
- If you had an action whose name actually ended with "Action" (e.g., MyAction.action) and passed in the action name as "MyActionAction", you must also remove the extra "Action":
  ```diff
  - const ac = nh.actionClientInterface('/turtle_shape', 'my_msgs/MyActionAction');
  + const ac = nh.actionClientInterface('/turtle_shape', 'my_msgs/MyAction');
  ```
- If your action type did not include the extra "Action", no changes are necessary:
  ```js
  const ac = nh.actionClientInterface('/turtle_shape', 'turtle_actionlib/Shape');
  ```

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,7 +92,7 @@ const client = nh.serviceClient('/add_two_ints', AddTwoInts);
 const nh = rosnodejs.nh;
 const as = new rosnodejs.ActionServer({
   nh,
-  type: 'turtle_actionlib/ShapeAction',
+  type: 'turtle_actionlib/Shape',
   actionServer: '/turtle_shape'
 });
 
@@ -102,7 +102,7 @@ as.on('goal', function (goal) {
 
 as.start();
 
-const ac = nh.actionClientInterface('/turtle_shape', 'turtle_actionlib/ShapeAction');
+const ac = nh.actionClientInterface('/turtle_shape', 'turtle_actionlib/Shape');
 ac.sendGoal({ goal: {edges: 3, radius: 1}});
 ```
 ## Run the turtlesim example
--- a/src/examples/turtle.js
+++ b/src/examples/turtle.js
@@ -77,7 +77,7 @@ rosnodejs.initNode('/my_node', {onTheFly: true}).then((rosNode) => {
       const shapeActionGoal = rosnodejs.require('turtle_actionlib').msg.ShapeActionGoal;
       const ac = rosnodejs.nh.actionClient(
         "/turtle_shape",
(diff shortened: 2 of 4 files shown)


jstnhuang:
Gah, looks like Travis failed for an unrelated reason, could you rerun that job?

chris-smith:
lgtm! Thanks!

PR #107, opened 2018-07-25
Add support for tcpNoDelay on subscribers

chris-smith:
We were missing the ability for users to specify `tcpNoDelay` on subscribers - this is the only way to make `roscpp` publishers set no delay on their socket.

Also does some cleanup (removing unused functions) and renaming.

### DIFF
--- a/src/lib/impl/PublisherImpl.js
+++ b/src/lib/impl/PublisherImpl.js
@@ -43,8 +43,7 @@ class PublisherImpl extends EventEmitter {
 
     this._tcpNoDelay =  !!options.tcpNoDelay;
 
-
-    if (options.queueSize) {
+    if (options.hasOwnProperty('queueSize')) {
       this._queueSize = options.queueSize;
     }
     else {
@@ -165,7 +164,7 @@ class PublisherImpl extends EventEmitter {
   }
 
   /**
-   * Check if this subscriber has been shutdown
+   * Check if this publisher has been shutdown
    * @returns {boolean}
    */
   isShutdown() {
@@ -242,10 +241,10 @@ class PublisherImpl extends EventEmitter {
   /**
    * Handles a new connection from a subscriber to this publisher's node.
    * Validates the connection header and sends a response header
-   * @param subscriber {Socket} client that sent the header
+   * @param socket {Socket} client that sent the header
    * @param header {Object} deserialized connection header.
    */
-  handleSubscriberConnection(subscriber, header) {
+  handleSubscriberConnection(socket, header) {
     let error = TcprosUtils.validateSubHeader(
       header, this.getTopic(), this.getType(),
       this._messageHandler.md5sum());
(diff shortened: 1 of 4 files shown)


src/lib/impl/PublisherImpl.js
```diff
@@ -43,7 +43,6 @@ class PublisherImpl extends EventEmitter {
 
     this._tcpNoDelay =  !!options.tcpNoDelay;
 
-
     if (options.queueSize) {
```
rethink-rlinsalata: options.queueSize doesn't allow 0 queueSize!

src/utils/tcpros_utils.js
```diff
@@ -28,10 +28,11 @@ const md5Prefix = 'md5sum=';
 const topicPrefix = 'topic=';
 const servicePrefix = 'service=';
 const typePrefix = 'type=';
-const latchingPrefix = 'latching=';
-const persistentPrefix = 'persistent=';
 const errorPrefix = 'error=';
 const messageDefinitionPrefix = 'message_definition=';
+const latchingField = 'latching=1';
+const persistentField = 'persistent=1';
+const tcpNoDelayField = 'tcp_nodelay=1'
```
rethink-rlinsalata: missing semi-colon =P
(i know, i know)

rethink-rlinsalata (changes requested):
Fix then ship it!

src/utils/tcpros_utils.js
```diff
@@ -28,10 +28,11 @@ const md5Prefix = 'md5sum=';
 const topicPrefix = 'topic=';
 const servicePrefix = 'service=';
 const typePrefix = 'type=';
-const latchingPrefix = 'latching=';
-const persistentPrefix = 'persistent=';
 const errorPrefix = 'error=';
```
rethink-rlinsalata, line 31: Do you use any variable naming conventions in rosnodejs for (actual) constants? Like all CAPS or something similar?
I don't think this is an issue at all here, especially since it is consistent and these are all specifically named as Field's and Prefix's, just curious,

src/lib/impl/PublisherImpl.js
chris-smith: :+1:

src/utils/tcpros_utils.js
chris-smith: good eye :eye:

src/lib/impl/PublisherImpl.js
rethink-rlinsalata: Just for the sake of git-commit historians, might want to mention that this fixes the queueSize=0 option (Infinite queueSize).

src/utils/tcpros_utils.js
chris-smith: I think I generally do CAPS_UNDERSCORE... this was one of the first things I wrote when I was still just messing around though so I probably wasn't thinking about it.

rethink-rlinsalata (approved):

PR #120, opened 2018-11-08
Fixes bug in on-the-fly message handling of unicode strings

### LINKED ISSUE #119: Unicode characters in strings causes fatal error (using on-the-fly messages)
At least when using on-the-fly message definitions, when publishing messages that include strings with Unicode characters this fatal error occurs:

```
RangeError: Index out of range
    at checkInt (buffer.js:1187:11)
    at Buffer.writeUInt32LE (buffer.js:1265:5)
    at /home/cfritz/tmp/out_of_bounds/node_modules/rosnodejs/dist/utils/messageGeneration/messages.js:545:16
    at Array.forEach (<anonymous>)
    at serializeInnerMessage (/home/cfritz/tmp/out_of_bounds/node_modules/rosnodejs/dist/utils/messageGeneration/messages.js:540:15)
    at Function.Message.serialize (/home/cfritz/tmp/out_of_bounds/node_modules/rosnodejs/dist/utils/messageGeneration/messages.js:488:5)
    at Object.serializeMessage (/home/cfritz/tmp/out_of_bounds/node_modules/rosnodejs/dist/utils/tcpros_utils.js:195:18)
    at msgQueue.forEach.msg (/home/cfritz/tmp/out_of_bounds/node_modules/rosnodejs/dist/lib/impl/PublisherImpl.js:224:43)
    at Array.forEach (<anonymous>)
    at PublisherImpl._handleMsgQueue (/home/cfritz/tmp/out_of_bounds/node_modules/rosnodejs/dist/lib/impl/PublisherImpl.js:219:16)
```

This is the minimal code I was able to reproduce this with:

```js
const rosnodejs = require('rosnodejs');

rosnodejs
  .initNode('/test', { onTheFly: true })
  .then((rosNode) => {

  const pub = rosNode.advertise('/rosout', 'rosgraph_msgs/Log', {});
  const Log = rosnodejs.require('rosgraph_msgs').msg.Log;

  const msg = '’'; // a unicode character

  pub.publish(new Log({
    level: 1,
    name: rosNode.nodeName,
    msg,
    file: '',
    function: 'test',
    line: 0,
    topics: []
  }));
});
```

chfritz:
Fixes #119.

Thanks for the pointer @chris-smith !

### DIFF
--- a/src/utils/messageGeneration/fields.js
+++ b/src/utils/messageGeneration/fields.js
@@ -283,7 +283,7 @@ fields.getPrimitiveSize = function(fieldType, fieldValue) {
   }
   else if (fieldType === 'string') {
     if (fieldValue !== undefined) {
-      fieldSize = fieldValue.length + 4;
+      fieldSize = Buffer.byteLength(fieldValue, 'utf-8') + 4;
     }
   }
   else if (fieldType === 'time') {
--- a/src/utils/messageGeneration/fields.js
+++ b/src/utils/messageGeneration/fields.js
@@ -283,7 +283,7 @@ fields.getPrimitiveSize = function(fieldType, fieldValue) {
   }
   else if (fieldType === 'string') {
     if (fieldValue !== undefined) {
-      fieldSize = Buffer.byteLength(fieldValue, 'utf-8') + 4;
+      fieldSize = Buffer.byteLength(fieldValue, 'utf8') + 4;
     }
   }
   else if (fieldType === 'time') {


src/utils/messageGeneration/fields.js
```diff
@@ -283,7 +283,7 @@ fields.getPrimitiveSize = function(fieldType, fieldValue) {
   }
   else if (fieldType === 'string') {
     if (fieldValue !== undefined) {
-      fieldSize = fieldValue.length + 4;
+      fieldSize = Buffer.byteLength(fieldValue, 'utf-8') + 4;
```
chris-smith: Minor, but should this be `'utf8'` instead of `'utf-8'` ([docs](https://nodejs.org/api/buffer.html#buffer_buffers_and_character_encodings))? It's unclear to me if this is an accepted alternative or if it's falling back to the default value (`'utf8'`) because it doesn't recognize the provided option.

chfritz: Good catch! Fixed, thanks!

chris-smith (approved):

PR #129, opened 2019-08-01
Update .travis.yml

chris-smith:


chris-smith:
This updates the gpg key (re [New GPG keys deployed](https://discourse.ros.org/t/new-gpg-keys-deployed-for-packages-ros-org/9454)) and migrates to xenial/kinetic from trusty/indigo which are EOL.

PR #130, opened 2019-08-22
Emit ActionServer updates

lukaszmitka:
Emit events when there is update on status, feedback or result.


chris-smith:
Is there a specific reason you want to do this through the `ActionClient` too, rather than using the `ActionClientInterface`? The `ActionClientInterface` basically just subscribes to all the topics and emits these events when the corresponding messages are received. The `ActionClient` class is meant to mimic `actionlib`.

lukaszmitka:
Thank you for clarification. Then this PR is not necessary.

PR #137, opened 2019-11-26
Fix regex for parsing url.

### LINKED ISSUE #136: Wrongly parsing host name
This lines process the regular expression in the wrong way in the file utils/network_utils.js:74 (called by lib/impl/SubscriberImpl.js:210):

`
> let regexStr = /(?:http:\/\/|rosrpc:\/\/)?([a-zA-Z\d\-.:]+):(\d+)/;
undefined
> regexStr
/(?:http:\/\/|rosrpc:\/\/)?([a-zA-Z\d\-.:]+):(\d+)/
> let uriString = 'http://weather_reporter:40779/'
undefined
> uriString
'http://weather_reporter:40779/'
> let match = uriString.match(regexStr);
undefined
> match[1]
'reporter'
> 
`

and I get
`
> uriString.match(regexStr2);
[
  'http://weather_reporter:40779',
  'weather_reporter',
  '40779',
  index: 0,
  input: 'http://weather_reporter:40779/',
  groups: undefined
]
> 
`

It should be weather_reporter. Apart from changing the name of my node, would you recommend how to fix it?

I had temporary edited the regular expression to:
`
/(?:http:\/\/|rosrpc:\/\/)?([a-zA-Z\d\-_.:]+):(\d+)/
`

Renato.

chfritz:
Added `_` to the valid characters in the hostname. Note that I also removed the `:`. I think that's only be allowed at the end (where we already have it), before the port.

Fixes #136.

### DIFF
--- a/src/utils/network_utils.js
+++ b/src/utils/network_utils.js
@@ -74,7 +74,7 @@ const NetworkUtils = {
   },
 
   getAddressAndPortFromUri(uriString) {
-    let regexStr = /(?:http:\/\/|rosrpc:\/\/)?([a-zA-Z\d\-.:]+):(\d+)/;
+    let regexStr = /(?:http:\/\/|rosrpc:\/\/)?([a-zA-Z\d\-_.]+):(\d+)/;
     let match = uriString.match(regexStr);
     if (match === null) {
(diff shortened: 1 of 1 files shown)


chris-smith:
Yep makes sense. Thanks!

PR #138, opened 2019-12-20
Udp Support

cicciodifranco:
Hello,
We implemented support to udp transport for publishers and subscribers
it was implemented as described [here](http://wiki.ros.org/ROS/UDPROS) but only opcode 0 and 1.
There are many things to implement / fix / clean 
- Support to opcode 2 and 3
- Remove unused function in src/utils/udpros_utils.js
- Destroy udp server socket after client disconnect

TCP is used as default if no options was provided

This is not a production version but useful for experimental tests...


src/lib/impl/PublisherImpl.js
```diff
@@ -224,12 +229,60 @@ class PublisherImpl extends EventEmitter {
           this._subClients[client].write(serializedMsg);
         });
 
+        Object.keys(this._udpSubClients).forEach((client) => {
```
chris-smith: can this all just be moved into a `_sendMsgToUdpClients` method?

src/lib/impl/SubscriberImpl.js
```diff
@@ -290,15 +327,109 @@ class SubscriberImpl extends EventEmitter {
...
+
+    socket.on('listening', () => {
+      const address = socket.address();
+      this._log.debug(`UDP socket bound: ${address.address}:${address.port}, Topic:${this.getTopic()}`);
+      
+    });
+
+    this._subServer = socket
+  
+    socket.bind(this._port);
```
chris-smith: Can this be moved into RosNode.js so that there's only one UDP server, mimicking `_setupTcprosServer`? Should be able to just pass 0 for port too in order to get a random available port.

src/lib/RosNode.js
```diff
@@ -109,11 +113,41 @@ class RosNode extends EventEmitter {
 
     return new Publisher(pubImpl);
   }
+  
+  /**
+   * get udp bound ports
+   * @returns [Number]
+   */
+  getBoundPorts(){
```
chris-smith: presumably could get rid of these methods if udp server set up is moved here

src/lib/impl/SubscriberImpl.js
```diff
@@ -290,15 +327,109 @@ class SubscriberImpl extends EventEmitter {
...
+      }
+    });
+
+    socket.on('listening', () => {
+      const address = socket.address();
+      this._log.debug(`UDP socket bound: ${address.address}:${address.port}, Topic:${this.getTopic()}`);
+      
+    });
+
+    this._subServer = socket
```
chris-smith: this server should be shutdown with the `_tcprosServer` and `_slaveApiServer` on shutdown too

chris-smith (changes requested):

cicciodifranco:
ok 
We will implement changes asap

src/lib/RosNode.js
```diff
@@ -109,11 +113,41 @@ class RosNode extends EventEmitter {
...
+  }
 
   subscribe(options, callback) {
     let topic = options.topic;
     let subImpl = this._subscribers[topic];
     if (!subImpl) {
+
       subImpl = new SubscriberImpl(options, this);
       this._subscribers[topic] = subImpl;
     }
```
cicciodifranco: Is it possible to subscribe several times with both udp and tcp?

src/lib/SlaveApiClient.js
```diff
@@ -35,6 +35,8 @@ class SlaveApiClient {
           reject(err, resp);
         }
         else {
+
+          //console.log(resp[2][5].toString())
```
chris-smith: delete?

src/lib/impl/SubscriberImpl.js
```diff
@@ -141,7 +214,7 @@ class SubscriberImpl extends EventEmitter {
 
     Object.keys(this._pubClients).forEach(this._disconnectClient.bind(this));
     Object.keys(this._pendingPubClients).forEach(this._disconnectClient.bind(this));
-
+    Object.keys(this)
```
chris-smith: this seems like it should be deleted?

src/lib/NodeHandle.js
```diff
@@ -124,6 +124,19 @@ class NodeHandle {
         options.typeClass = type;
         options.type = type.datatype();
       }
+      options.tcp = options.tcp === undefined ? false : options.tcp
```
chris-smith: I'm wondering if it would be better if we either mirrored roscpp's options here, so the options could include a `transports` list and we default to TCPROS if nothing is specified or just include the `udp` flag option here.

Mostly not sure about the combination of `udp`, `tcp`, and `udpFirst` to order them...

src/lib/impl/PublisherImpl.js
```diff
@@ -86,6 +90,7 @@ class PublisherImpl extends EventEmitter {
 
     this._state = REGISTERING;
     this._register();
+    this.udpSocket = Udp.createSocket('udp4')
```
chris-smith: should we only create this socket if we add a udp subscriber?

chris-smith (commented):
mostly just the 2 minor comments about the options and when to create a udp socket

chris-smith (approved):
Changes look good. Thanks!

cicciodifranco:
Did you published these changes in `npm`?

PR #139, opened 2020-01-10
Fix error on action cancellation

QuentinLB:
By using `_clientInterface` instead of `_actionClient` we prevent the following error : "TypeError: Cannot read property 'cancel' of undefined" caused by the fact that `_actionClient` is never properly assigned.

### DIFF
--- a/src/actions/ClientGoalHandle.js
+++ b/src/actions/ClientGoalHandle.js
@@ -83,7 +83,7 @@ class ClientGoalHandle extends EventEmitter {
         return;
     }
 
-    this._actionClient.cancel(this._goal.goal_id.id, { secs: 0, nsecs: 0 });
+    this._clientInterface.cancel(this._goal.goal_id.id, { secs: 0, nsecs: 0 });
     this._transition(CommState.WAITING_FOR_CANCEL_ACK);
   }
 


chris-smith:
Thanks for the fix!

QuentinLB:
@chfritz 

Is it possible to publish the head of kinetic-devel as a 3.0.2 version on npm to get the last PR ? I tried pulling like this : `"rosnodejs": "github:RethinkRobotics-opensource/rosnodejs#kinetic-devel",` but I realized that the project needed a build step before being usable as a dependency. Publishing a new version would avoid creating a fork and another npm repo (under another scope).

Thanks

chfritz:
I think this is a question for @chris-smith . However, if you can't wait, then you can definitely use a fork as dependency by running `npm run compile` in it and committing the resulting `dist/` folder. So at least you can avoid creating another npm repo. That's what I've done in the past when necessary, so you can follow [my example](https://github.com/chfritz/rosjs/tree/with_dist_2.1.1) if you want.

chris-smith:
I should be able to push an update tomorrow.

QuentinLB:
@chfritz Sorry for the ping it was meant for Chris, nevertheless thanks for your answer and the example.

@chris-smith Thank you :+1:

Issue #118, opened 2018-10-22
bad performance when deserialize long buildin-type array


panshengjie:
rosnodejs deserialize buildin-type array data oneByone using deserializePrimitive(),   can we use typeArray instead?

chris-smith:
As far as I'm aware. The uint8 arrays for pre-generated messages are already specially deserialized so we'd just be extending the case.

chfritz:
Is this a duplicate of https://github.com/RethinkRobotics-opensource/rosnodejs/issues/117?

panshengjie:
@chfritz  not the same.  I thought we should use typeArray everywhere  supported,  great performance improve you know.  But this may lead to compatibility issues.

chris-smith:
A few things I was trying to determine (and hadn't managed to yet)...

- are TypedArrays strictly a superset of normal Arrays? Basically, that they officially supported all the same properties/methods as normal arrays but had additional ones.
- how does this impact older versions of Node that we currently support? Are there babel plugins for it?
- Do we need some flag to indicate that generated messages should deserialize to TypedArrays instead of normal arrays? This probably depends on the previous 2 answers...

We'd also just need some perf tests to show the improvement we expect (mostly to guard against bugs in implementation, assuming TypedArrays would be a performance improvement in general).

panshengjie:
it's quite sure that, TypedArray is not a  superset of normal Arrays, which is introduced for binary operation that nodejs lacks. Replace all arrays with typeArray is impossiable, especially object-arrays.
 
a good choice I think about is introduce a flag to let programmer manually handle heavy binary datas like image/video/sound

besides, heavy object-arrays are still tough , typically laserscan/pointclouds in ROS.

chris-smith:
I started looking into doing this but discovered that the TypedArray views require that the byte offset be a multiple of the type size (see: [StackOverflow](https://stackoverflow.com/questions/15417310/why-typed-array-constructors-require-offset-to-be-multiple-of-underlying-type-si)). In specific circumstances (e.g for certain message definitions) it can be guaranteed that this behavior will hold, but turning on TypedArray deserialization for specific topics or messages and not others would probably be a more invasive change than what I'd been imagining...

The general way around the byte-offset issue for ArrayBuffers seems to be using a DataView, but the API there looks like we'd end up with basically the same code we currently have now using Node.js Buffers (which may just be a DataView at this point??).

[...]

PR #153, opened 2020-07-01
little changes to speedup topic listen, especially custom msg array

panshengjie:
I have done some test,  result in 5x faster around, and some mem reduce.

when listen to /tf  /scan /imageStream ...  topics with very high frequency, these changes are helpful  #118

### DIFF
--- a/src/utils/messageGeneration/messages.js
+++ b/src/utils/messageGeneration/messages.js
@@ -603,11 +603,7 @@ function deserializeInnerMessage(spec, message, buffer, bufferOffset) {
         bufferOffset[0] += 4; // only for variable length arrays
       }
 
-      let ArrayMsgClass;
       let isPrimitive = field.isBuiltin;
-      if (!isPrimitive) {
-        ArrayMsgClass = getMessageFromRegistry(arrayType, 'msg');
-      }
 
       for (var i = 0; i < arraySize; i++) {
         if (isPrimitive) {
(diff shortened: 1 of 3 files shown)


src/utils/messageGeneration/messages.js
```diff
@@ -603,11 +603,7 @@ function deserializeInnerMessage(spec, message, buffer, bufferOffset) {
         bufferOffset[0] += 4; // only for variable length arrays
       }
 
-      let ArrayMsgClass;
       let isPrimitive = field.isBuiltin;
-      if (!isPrimitive) {
-        ArrayMsgClass = getMessageFromRegistry(arrayType, 'msg');
```
chris-smith, line 609: I'm not sure about getting rid of these classes during deserialization, especially since they seem to be class instances otherwise. I'm assuming the main issue here is that we're doing this lookup on every deserialization, so I'd prefer to find a way to only do this once.

Not sure what @chfritz thinks as he has more experience using the on-the-fly messages.

panshengjie: yes, I got the same way at first time, but then I found this lookup means nothing, the spec will be got in line 631 spec.getMsgSpecForType(field.baseType), and arrayMessage that passed in , works as empty container.

chris-smith: Any clients that are relying on the deserialized format being `instanceof Message` won't be able to... I'd guess this isn't too widespread, but it's a little odd since we do it on [initialization](https://github.com/RethinkRobotics-opensource/rosnodejs/blob/kinetic-devel/src/utils/messageGeneration/messages.js#L445) of messages.

panshengjie: if we new the msg, we do know the msg type already.  I think all we do is to set/get data, like  `msg.data` .    and we still able to `instanceof Message` for whole msg, except child msg

chris-smith: yes it's a pretty minor case...

I think I'm ok to leave this as is for now. We can move back to having `Message` instances if needed, but I think we can do it in a more optimal way.

PR #158, opened 2020-08-11
Can't await .initNode in modern node.

JamesNewton:


chris-smith:
Do you mean because there's no top level await?

JamesNewton:
When I tried running the code sample as provided, it failed for that reason.

Issue #127, opened 2019-06-04
can't install because of chfritz/node-xmlrpc


ValeryVS:
`chfritz/node-xmlrpc` dependency can not be installed without ssh access using node>=8
For me `npm install` hanged.
Can you publish `node-xmlrpc` or set dependency explicitly with https protocol?

chris-smith:
Do you explicitly _not_ have ssh access for some reason? Does this resolve your issue https://docs.npmjs.com/common-errors#npm-only-uses-git-and-sshgit-urls-for-github-repos-breaking-proxies?

ValeryVS:
Yes, I haven't.
I definitely solve this with `git config url."https://".insteadOf git://`, but this workaround is quite hacky.

I wont use https every time, so `url."https://".insteadOf` should be set before every `npm install`, and deleted after every `npm install`. I know, this can be done with `preinstall` and `postinstall` scripts with npm>=5. And node@6, where `preinstall` is broken in npm@3, use https by default.

If new xmlrpc package could be published, it would be much easier.

PR #159, opened 2020-09-07
TF topic check, add message size and node uri, fix Buffer deprecation

chfritz:
This PR addresses one issue and one warning and adds two (minor) features:

- Issue: The strict topic type checking doesn't work with the `/tf` topic which is sometimes `tf/tfMessage` (gazebo) and sometimes `tf2_msgs/TFMessage` (seemingly everywhere else). The md5 is still the same though because the definition of these two types is actually the same. Apparently other ROS tools are not bothered by this mix-and-match of types on the same topic, so I think it's OK to omit the topic type check.

- Warning: Addressed this warning we were getting when starting:
  > (node:21200) [DEP0005] DeprecationWarning: Buffer() is deprecated due to security and usability issues. Please use the Buffer.alloc(), Buffer.allocUnsafe(), or Buffer.from() methods instead.
  
  Mostly using Buffer.allocUnsafe now, since we'll fill these buffers anyways, which might also be slightly faster.

- Features: provide the subscriber callback also with the message size in bytes as well as the publishers nodeUri. This is backwards compatible, since they are just additional arguments to the callback, i.e., will be ignored by current code. This information is useful mostly for informational purposes but also for filtering (selective listening). I use this in [ros-blessed](https://www.npmjs.com/package/ros-blessed).

-----
**Update**

Now also addresses https://github.com/RethinkRobotics-opensource/rosnodejs/issues/127 by using a newly published, separate npm package (https://www.npmjs.com/package/xmlrpc-rosnodejs) instead of depending on the github repo, as suggested in the ticket.

### DIFF
--- a/src/utils/tcpros_utils.js
+++ b/src/utils/tcpros_utils.js
@@ -41,7 +41,7 @@ function serializeStringFields(fields) {
   fields.forEach((field) => {
     length += (Buffer.byteLength(field) + 4);
   });
-  let buffer = new Buffer(4 + length);
+  let buffer = Buffer.allocUnsafe(4 + length);
   let offset = base_serializers.uint32(length, buffer, 0);
 
   fields.forEach((field) => {
@@ -214,11 +214,11 @@ let TcprosUtils = {
     let msgBuffer;
     let offset = 0;
     if (prependMessageLength) {
-      msgBuffer = new Buffer(msgSize + 4);
+      msgBuffer = Buffer.allocUnsafe(msgSize + 4);
       offset = base_serializers.uint32(msgSize, msgBuffer, 0);
     }
     else {
-      msgBuffer = new Buffer(msgSize);
+      msgBuffer = Buffer.allocUnsafe(msgSize);
     }
 
     MessageClass.serialize(message, msgBuffer, offset);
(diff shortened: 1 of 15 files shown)


chris-smith:
Awesome! I think this looks good but can't do more than glance right now - should have time next week to look more.

src/utils/tcpros_utils.js
```diff
@@ -181,10 +181,6 @@ let TcprosUtils = {
     else if (!header.hasOwnProperty('md5sum')) {
       return this.serializeString('Connection header missing expected field [md5sum]');
     }
-    // rostopic will send '*' for some commands (hz)
```
chris-smith, line 184: can you just add a comment here with the TF explanation for why we don't check the type field?

chris-smith (changes requested):
Looks good!

src/utils/tcpros_utils.js
chfritz: Good idea. Done.

PS: note that I squashed the commit where I added the note into the commit for the above change and force-pushed. Just in case you were wondering why there aren't any new commits.

chris-smith: thanks

chris-smith (approved):

PR #164, opened 2020-10-06
Adds new function getServiceType to NodeHandle: looks up type from server

chfritz:
As explained in [ros_comm](https://github.com/ros/ros_comm/blob/6292d54dc14395531bffb2e165f3954fb0ef2c34/tools/rosservice/src/rosservice/__init__.py#L94-L98):
> Services currently do not declare their type with the master, so instead we probe the service for its headers.

The added function does just that. Test included.

Mini-example:

```lang-js
const type = await rn.getServiceHeader('/rosout/get_loggers');
console.log(type)
```

Output:
```
{
  callerid: '/rosout',
  md5sum: '32e97e85527d4678a8f9279894bb64b0',
  request_type: 'roscpp/GetLoggersRequest',
  response_type: 'roscpp/GetLoggersResponse',
  type: 'roscpp/GetLoggers'
}
```

### DIFF
--- a/src/lib/RosNode.js
+++ b/src/lib/RosNode.js
@@ -299,6 +299,41 @@ class RosNode extends EventEmitter {
     return this._masterApi.getSystemState(this._nodeName, options);
   }
 
+  /** Services currently do not declare their type with the master, so instead
+    we probe the service for its headers. Just like in
+  https://github.com/ros/ros_comm/blob/6292d54dc14395531bffb2e165f3954fb0ef2c34/tools/rosservice/src/rosservice/__init__.py#L94-L98
+  */
+  getServiceHeader(serviceName) {
+    return new Promise((resolve, reject) => {
+
+      this.lookupService(serviceName).then((resp) => {
+        const serviceUri = resp[2];
+        const serviceHost = NetworkUtils.getAddressAndPortFromUri(serviceUri);
+
+        const client = net.connect(serviceHost, () => {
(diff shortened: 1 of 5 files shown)


src/lib/RosNode.js
```diff
@@ -299,6 +299,41 @@ class RosNode extends EventEmitter {
     return this._masterApi.getSystemState(this._nodeName, options);
   }
 
+  /** Services currently do not declare their type with the master, so instead
+    we probe the service for its headers. Just like in
+  https://github.com/ros/ros_comm/blob/6292d54dc14395531bffb2e165f3954fb0ef2c34/tools/rosservice/src/rosservice/__init__.py#L94-L98
+  */
+  getServiceType(serviceName) {
```
chris-smith: should this be called `getServiceHeaders`? `ros_comm` has both and `getServiceType` is a wrapper around `getServiceHeaders`

chfritz: The different between the two in ros_comm seems to be that `get_service_type` takes just the service name as argument and looks up the node whereas `get_service_headers` requires the service URI as argument. So in that analogy `getServiceType` would seem to be the right name to use here.

chris-smith: I was mostly looking at the return type - `get_service_headers` returns headers (like this function) while `get_service_type` returns only the type field of the header.

chfritz: Good point, yes. In that case I agree. Done.

PR #167, opened 2020-10-15
Fix typo (parameter name didn't match usage in function)

derrell:
At the same time, update other location which used same parameter
name, to match NodeJS standards as well as matching the names used in
the roscon 2017 presentation.


chris-smith:
Thanks for the catch!

PR #170, opened 2021-03-09
fix int64 endian BUG

panshengjie:
tested


chris-smith:
lgtm - thanks! I filed a ticket to fix the tests that let this get through.

PR #181, opened 2021-07-28
UDP bugfix

cicciodifranco:
This pr fix some bugs that we found in UDP messaging:

- in [UDPROS](http://wiki.ros.org/ROS/UDPROS) `msgID` is `uint8` so after 255 messages sent the node throw an exception
- `host` sent to the publisher during handshake was wrong. (the host sent was the ip sent by the publisher so this made impossible to work in a multi host configuration)


src/lib/impl/PublisherImpl.js
```diff
@@ -241,7 +241,7 @@ class PublisherImpl extends EventEmitter {
         if (this.getLatching()) {
           this._lastSentMsg = serializedMsg;
         }
-        msgCount++;
+        msgCount = msgCount > 253 ? 0 : msgCount++;
```
chris-smith: why does this rollover at 253?

cicciodifranco: Using this expression the maximum value that we can reach is 254. I'm updating the expression to reach 255.
Btw the expression should be ( post increment will leave `msgCount` to 0 )
```
msgCount = msgCount > 254 ? 0 : ++msgCount;
```

gtropea: > why does this rollover at 253?

the rollover at 253 (better, at 254, see next commits) is necessary because in the udp header we use, the msgId is a uint8: see src/utils/udpros_utils.js line 283

gtropea:
We have now extensively tested the UDP feature in a heterogeneous setup, i.e. rosnodejs talking to roscpp in UDP (both of them subscriber and publisher) (in addition to rosnodejs-to-rosnodejs), and it works like a charm! please @chris-smith review these latest fixes!

chris-smith:
thanks for the contribution!

PR #205, opened 2024-10-19
Fix backticks in message definitions, throw better error on missing dependencies.

### LINKED ISSUE #204: Backticks in msg definition comments result in invalid JS syntax
In working with a user of [Transitive](https://github.com/transitiverobotics/transitive), I discovered that backticks in comments in `.msg` files result in auto-generated JS code that is invalid (the included backticks end the multi-line string we generated that itself uses backticks as delimiter). Unfortunately, there are even official ROS packages like [GridMap](https://docs.ros.org/en/jade/api/grid_map_msgs/html/msg/GridMap.html) that use backticks in their comments.

The resulting error when trying to load those files is (here for GridMapInfo):
```
[2024-10-18T22:44:40.681Z main error] **** Caught exception: SyntaxError: Unexpected identifier 'header': /opt/ros/noetic/share/gennodejs/ros/grid_map_msgs/msg/GridMapInfo.js:126
    # Pose of the grid map center in the frame defined in `header` [m].
                                                           ^^^^^^

SyntaxError: Unexpected identifier 'header'
```

The good news is that it is easy to fix, and I already have a fix ready [here](https://github.com/chfritz/rosjs/commit/7fa86b859142f9766ca962f2e0e9a89b98c4f506).

@chris-smith I'm not certain how gennodejs works these days. Is it still using [this completely separate python implementation](https://github.com/RethinkRobotics-opensource/gennodejs/blob/kinetic-devel/src/gennodejs/generate.py) or does it use the message generation from this repo by now? If the latter, then the above fixes would be easy to apply. If it's a completely separate implemetation in Python we would need to fix it there, too.

I'll create a PR for the node.js version for now.

chfritz:
Fixes #204 -- at least for this repo.

- [Message gen: escape backticks in comments](https://github.com/RethinkRobotics-opensource/rosnodejs/commit/21c3e46d15808f29ad6b8cd2bca31c9cc9ff3e6c) 
  Until now, backticks ("`") in comments in message definitions will lead to an exception when the generated message is loaded, because the generated JS code is syntactically invalid.

- [Message generation: indicate name of missing dependencies](https://github.com/RethinkRobotics-opensource/rosnodejs/commit/9966bcabd7adf19492cfae32a92e51e8a7ba9c21) 
  Previously, if a custom message definition referred to another package that was missing, the error was not giving any clues as to where the problem occurred, i.e., which packages are involved. This change fixes that, making it much easy to debug broken dependencies.

### DIFF
--- a/src/utils/messageGeneration/MessageLoader.js
+++ b/src/utils/messageGeneration/MessageLoader.js
@@ -394,6 +394,9 @@ class MessageManager {
       const depIndex = dependencyChain.indexOf(depName);
       if (depIndex === -1) {
         // this dependency is not yet in the list anywhere
+        if (!packageCache[depName]) {
+          throw Error(`Missing dependency ${depName}, required by ${packageName}`);
+        }
         const insertionIndex = this._recurseDependencyChain(dependencyChain, depName);
         if (insertionIndex > maxInsertionIndex) {
           maxInsertionIndex = insertionIndex;
--- a/src/utils/messageGeneration/MessageSpec.js
+++ b/src/utils/messageGeneration/MessageSpec.js
@@ -264,13 +264,17 @@ class RosMsgSpec {
 
     const deps = this.getFullDependencies();
     const sep = '='.repeat(80);
-    w.write(this.fileContents.trim())
+
+    // escape backticks in string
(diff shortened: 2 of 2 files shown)


chris-smith:
Thanks!

PR #207, opened 2026-06-25
Respond to shutdown

ablakey-locusrobotics:


ablakey-locusrobotics:
Apologies. Wrong target/branch.

For posterity, there seems to be an issue with shutdown where it doesn't respond to the caller, so `rosnode kill` hangs indefinitely.  This fixes it:


```typescript
  _handleShutdown(err, params, callback) {
    let caller = params[0];
    this._log.warn('Received shutdown command from ' + caller);
    // Acknowledge the caller before shutting down. The xmlrpc server only writes
    // an HTTP response when this callback fires, so failing to call it causes
    // `rosnode kill` to hang waiting for an acknowledgement that never arrives.
    callback(null, [1, 'shutdown received', 0]);
    return this.shutdown();
  }
```

